PR #1, opened 2024-09-20
Add RPC Server for multiplexing

avalonche:


src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+async fn main() -> Result<()> {
+    let args: Args = Args::parse();
+
+    // Initialize logging
+    tracing_subscriber::fmt()
+        .with_env_filter(EnvFilter::new(args.log_level.to_string())) // Set the log level
+        .init();
+
+    // Initialize the local execution engine client
+    let mut headers = HeaderMap::new();
```
ferranbt, line 62: Did you try this part? I think for making http client requests you need to do [sign the token](https://github.com/ethereum-optimism/op-geth/blob/5f7ebba8a124ae87225f81a1a9c827f8a534f2b7/node/jwt_auth.go#L33) not only include it on the header.

src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    let eth_engine_api = EthEngineApi::new(Arc::new(l2_client), Arc::new(builder_client));
+    let mut module = RpcModule::new(());
+    module
+        .merge(eth_engine_api.into_rpc())
+        .map_err(|e| Error::InitRPCServerError(e.to_string()))?;
+
+    // server setup
+    info!("Starting server on :{}", args.rpc_port);
+    let server = Server::builder()
```
ferranbt, line 88: Because of time constraints in the Go version I had to explicitly write down all the proxy functions myself being called by op-node. But, for the majority of them, we just want to relay the full content as is to the op-geth node.
A better approach, if we have access to some low-level opts of the http/Json-RPC lib would be to by default relay any request and only intercept and modify the endpoints we need.

ferranbt (approved):

src/server.rs
```diff
@@ -0,0 +1,213 @@
...
+    async fn get_proof(
+        &self,
+        address: Address,
+        keys: Vec<JsonStorageKey>,
+        block_number: Option<BlockId>,
+    ) -> RpcResult<EIP1186AccountProofResponse>;
+}
+
+#[rpc(server, client, namespace = "engine")]
+pub trait EngineApi {
```
liamaharon, line 48: Can reth traits be used?

src/main.rs
```diff
@@ -0,0 +1,100 @@
+use std::net::SocketAddr;
```
liamaharon, line 1: Small crate doc up here would be nice.

src/server.rs
ferranbt: This is a subset of the Reth API right? If we import the whole Eth API we have to implement all the methods?

PR #18, opened 2024-10-10
Osiris/reth compatibility

0xOsiris:
**Overview**
- This PR removes the assumption that the build process id of the builder matches the build process id of the l2 client. 

When testing with a reth builder the process id's didn't match which downstream caused issues in `get_payload_v3` when trying to grab the payload from the builder. 

**Solution**
- Ephemerally map the build process id of the l2 client to the build process id of the builder in `fork_choice_update_v3` when a build process has been started.
- The downside here is we now need to wait for both futures to resolve in `fork_choice_update_v3`

Note: This has been load tested with both a Reth builder, and a op-geth builder in a rollup devnet.


0xOsiris:
Closing as this is being addressed: https://github.com/paradigmxyz/reth/pull/11730

Issue #14, opened 2024-10-09
Multiplex `eth_sendRawTransaction` to the builder.


ferranbt:
Rollup-boost should multiplex the transactions sent with `eth_sendRawTransaction` to the builder too. Enable this behaviour with a flag.

avalonche:
this is easier to do at the rpc endpoint level via nginx than natively inside rollup-boost due to restrictions in the jsonrpcsee library.

ferranbt:
But with nginx you are adding an extra service to the mix, in this case the platform would do it natively.

Also, if `rollup-boost` had access to the transactions it could keep track on whether the builder is censoring them or not and act accordingly, for example, adding a few blocks from the proposer op-geth.

PR #22, opened 2024-10-25
Proxy eth_sendRawTransaction to builder

avalonche:
Resolves https://github.com/flashbots/rollup-boost/issues/14

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -9,6 +9,8 @@ use std::{future::Future, pin::Pin};
 use tower::{Layer, Service};
 use tracing::debug;
 
+const PROXY_METHODS: [&str; 2] = ["engine_", "eth_sendRawTransaction"];
+
 #[derive(Debug, Clone)]
 pub struct ProxyLayer {
     target_url: Uri,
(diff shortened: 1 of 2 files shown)


src/proxy.rs
```diff
@@ -84,7 +86,7 @@ where
                 message = "received json rpc request for",
                 method = method.method
             );
-            if method.method.starts_with("engine_") {
+            if PROXY_METHODS.iter().any(|&m| method.method.starts_with(m)) {
```
ferranbt, line 89: Does this mean that eth_ methods are not proxied? There are 3 eth_ methods that are also used in rollup-boost.

src/proxy.rs
```diff
@@ -9,6 +9,8 @@ use std::{future::Future, pin::Pin};
 use tower::{Layer, Service};
 use tracing::debug;
 
+const PROXY_METHODS: [&str; 2] = ["engine_", "eth_sendRawTransaction"];
```
ferranbt, line 12:
```suggestion
const MULTIPLEX_METHODS: [&str; 2] = ["engine_", "eth_sendRawTransaction"];
```
I would prefer to call it multiplex since it is the name we have been using so far for these methods.

ferranbt (approved):

PR #36, opened 2024-11-11
Add eth_sendBundle to multiplex to builder

avalonche:


ferranbt:
This would make sense if both the fallback and the external builder require to have access to the bundles. But, bundles are specific entities from the external builder so I am not sure if it is worth it to increase the scope of Rollup-boost for this.

avalonche:
do you suggest it as a method implemented in server.rs that is only forwarded to the builder?

dmarzzz:
I agree that for now we should have the sequencer's RPC proxy forward bundles until we have a need for this in rollup boost

PR #41, opened 2024-12-05
Add defaults for jwt token and l2 url

ferranbt:
This PR adds default values for `l2-url` (default authrpc address).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -31,14 +31,10 @@ mod server;
 
 #[derive(Parser, Debug)]
 #[clap(author, version, about)]
-#[clap(group(ArgGroup::new("jwt").multiple(false).args(&["jwt_token", "jwt_path"])))]
+#[clap(group(ArgGroup::new("jwt").required(true).multiple(false).args(&["jwt_token", "jwt_path"])))]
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -31,10 +31,14 @@ mod server;
...
 #[clap(author, version, about)]
-#[clap(group(ArgGroup::new("jwt").required(true).multiple(false).args(&["jwt_token", "jwt_path"])))]
+#[clap(group(ArgGroup::new("jwt").multiple(false).args(&["jwt_token", "jwt_path"])))]
 struct Args {
     /// JWT token for authentication
-    #[arg(long, env)]
+    #[arg(
+        long,
+        env,
+        default_value = "688f5d737bad920bdfb2fc2f488d6b6209eebda1dae949a8de91398d932c517a"
```
avalonche, line 40: I don't think there should be a default value here as the jwt token varies widely across deployments. This value is already in the .env.example file

ferranbt: I did not know that you could use env for this too. I will remove this one then.

avalonche (approved):

PR #47, opened 2025-01-14
chore: local -> external builder mapping

### LINKED ISSUE #46: Map payload_id of local builder to payload_ids of external builders
Removes the assumption that local builder should have the same payload id as external builders

namn-grg:
# Overview
This PR adds payload ID mapping functionality to track and map between local and external payload IDs when interacting with the builder node. This ensures proper correlation between payloads across different nodes.

Key changes:
- Added `local_to_external_payload_ids` cache to `PayloadTraceContext` to store mappings
- Added methods to store and retrieve payload ID mappings
- Updated `get_payload_v3` to use external payload ID when requesting from builder
- Added payload ID mapping in `fork_choice_updated_v3` response handling
- Added logging improvements to track both local and external payload IDs

The changes help maintain proper tracing and correlation between local execution engine payloads and external builder payloads, improving debuggability and reliability of the system.

fixes #46

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -526,6 +526,7 @@ mod tests {
     use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
     use jsonrpsee::server::{ServerBuilder, ServerHandle};
     use jsonrpsee::RpcModule;
+    use tokio::time::sleep;
 
     const L2_ADDR: &str = "0.0.0.0:8554";
     const BUILDER_ADDR: &str = "0.0.0.0:8555";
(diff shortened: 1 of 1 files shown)


namn-grg:
@avalonche can you pls check if the logic makes sense?
Its currently failing failing test
```
'server::tests::test_server' panicked at src/server.rs:760:9:
assertion `left == right` failed
  left: 0
 right: 1
 ```

**Comment by dmarzzz** [comment]
thanks for the PR! will review soon!

File: src/server.rs [hunk 1]
```diff
@@ -263,23 +303,41 @@ where
...
+                    .client
+                    .fork_choice_updated_v3(fork_choice_state, attr)
+                    .await
+                {
+                    Ok(response) => {
+                        let external_payload_id = response.payload_id;
+                        if let (Some(local_id), Some(external_id)) =
+                            (local_payload_id, external_payload_id)
+                        {
+                            payload_trace_context
```
avalonche, line 319: one potential optimismization here is that when the local_id == external_id it does not need the mapping and just default to the payload id given by the CL when get payload is called with an empty mapping

src/server.rs
```diff
@@ -235,18 +270,23 @@ where
                     3,
                 )
                 .unwrap();
-                let payload_id = builder_attrs.payload_id();
+                let local_payload_id = builder_attrs.payload_id();
```
avalonche, line 273: can just use `l2_response.payload_id` here

avalonche (approved):
LGTM, it'll be great if you can add some tests asserting when
1) payload ids are the same
2) payload ids are different
and check that the correct payload id is used. 
Also there seems to be lint issues

namn-grg:
Thanks for the review @avalonche . Addressed it!

src/server.rs
namn-grg: good point!

avalonche:
I've fixed some failing tests here

namn-grg:
Thanks! I see it was about running the tests sequentially. Personally not a fan of having all of them under `test_server` just looks dirty. Can be considered to use `serial_test` for it.

Issue #48, opened 2025-01-17
DA throttling


dmarzzz:
https://discord.com/channels/1244729134312198194/1294422364473528331/1327389163045257249

>Hey guy's wondering if any thought has been put into how to handle DA throttling within the block builder after the 1.9.5 batcher release. The op-batcher in this release will send a miner_setMaxDASize to the sequencing execution client to impose a throttle on tx size/max block size until the max pending queue bytes in the batcher is reduced below a configurable threshold. With an external block builder leveraging rollup-boost, the block builder will not be aware of the DA throttle imposed on the payloads, and will continue to build payloads at the standard gas limit accepted by the network. This opens up the chain to a spam attack. 

> I think there's two options here, wondering if anyone has a recommended solution:
Upstream changes to the batcher that multiplexes the miner_setMaxDASize call to the sequencer/(optional list of builder clients)
>Place a Proxy service in front of the batcher to multiplex the miner_setMaxDASize call to the sequencer/(optional list of builder clients) e.g. proxyd, or similar service to rollup-boost

>It seems like it might be advantageous to have this directly built into the batcher for chains running the rollup-boost stack, as to minimize complexity for the chain operator. It would be a fairly minor addition in the batcher.

ferranbt:
Rollup-boost includes a proxy service that relays certain RPC calls to both backends (e.g. `eth_sendRawTransaction`).

We could whitelist also the `miner_setMaxDASize` and other RPC methods required by op-batcher and have op-batcher connect with Rollup-boost directly instead of the sequencer EL node.

Note, we have to check exactly which RPC endpoints is op-batcher calling in the EL node, can we proxy all the methods? do some methods require some return parameters? do Do some calls require that only the EL node receives them?

0xKitsune:
> Note, we have to check exactly which RPC endpoints is op-batcher calling in the EL node, can we proxy all the methods? do some methods require some return parameters? do Do some calls require that only the EL node receives them?

The batcher calls the EL [within the main loop](https://github.com/ethereum-optimism/optimism/blob/develop/op-batcher/batcher/driver.go#L467-L474) to request blocks from the unsafe head. [Here is the actual call](https://github.com/ethereum-optimism/optimism/blob/develop/op-batcher/batcher/driver.go#L292) which is using `eth_getBlockByNumber` under the hood. Since a block is returned here, does it makes sense to forward this request to the sequencer's default execution client?


[...]

PR #60, opened 2025-01-27
feat(da-throttling): Forward `miner_` requests to the builder

0xKitsune:
ref #48 

This PR introduces logic to support DA throttling in the builder by forwarding `miner_` requests to the builder, following the same convention in the `EthApiServer` implementation. 

Feel free to let me know any feedback on the approach and if you have a preference on how to extend or amend the current test harness to best test this feature.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -216,7 +216,7 @@ where
 {
     async fn set_max_da_size(&self, bytes: Bytes) -> RpcResult<bool> {
         debug!(
-            message = "received set_max_da_size",
+            message = "received miner_setMaxDASize",
             "bytes_len" = bytes.len()
         );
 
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -116,6 +116,21 @@ pub trait EthApi {
     async fn send_raw_transaction(&self, bytes: Bytes) -> RpcResult<B256>;
 }
 
+#[rpc(server, client, namespace = "miner")]
+pub trait MinerApi {
+    #[method(name = "setMaxDASize")]
+    async fn set_max_da_size(&self, bytes: Bytes) -> RpcResult<bool>;
```
0xOsiris, line 122: Are we sure this is the correct selector?

I think it should be

```rust
async fn set_max_da_size(&self, max_tx_size: u64, max_block_size: u64) -> RpcResult<bool>;
```

0xKitsune: Thanks! I will update shortly.

Issue #65, opened 2025-01-29
Bug: Proxy Layer is forwarding all requests to the Auth RPC URI of the L2 execution engine


0xOsiris:
All requests hitting the Proxy layer are [forwarded to the auth RPC URI ](https://github.com/flashbots/rollup-boost/blob/81a921666d58ce59bc7f6c41b3dec9f2bbb6e407/src/proxy.rs#L102)of the L2 execution engine. However not all requests hitting `rollup-boost` are Admin RPC requests. In fact most will not be. 

Requests in the proxy layer need to be mapped to the correct target URI based on the API they are targeting on the execution engine.

Remediation:
- Rollup Boost needs to know the admin RPC port, and the http api port of both `op-geth`, and the builder(s)
- Rollup Boost at the proxy layer needs to determine the correct destination of the request based on the namespace of the request mapped to the correct port. 
- This also applies to the RPC server - Requests need to be mapped to the correct URI based on the target API

dmarzzz:
thanks for opening this!

ferranbt:
Thank you for opening this issue!

> However not all requests hitting rollup-boost are Admin RPC requests

What do you mean by Admin RPC requests? (engine_ requests, eth_ requests?). At least for the non-batcher use case, the op-node uses the authrpc endpoint for both engine and eth requests.

0xOsiris:
> Thank you for opening this issue!
> 
> > However not all requests hitting rollup-boost are Admin RPC requests
> 
> What do you mean by Admin RPC requests? (engine_ requests, eth_ requests?). At least for the non-batcher use case, the op-node uses the authrpc endpoint for both engine and eth requests.

Can we assume the request is coming from `op-node` at the Proxy layer though. `op-node` will never be sending `eth_sendRawTransaction` to the Proxy Layer. Communication over the `EthApi` will likely be coming from a Tx Proxy over the http API. 

The `op-batcher` will always be sending requests to the `EthApi`, and `MinerApi` over non auth RPC. 

The same can be said for `op-conductor`.

I think we need to be able to differentiate the target URI based on the format of the request in the Proxy Layer. We can't assume we are sending every request to the Auth RPC server on the L2 EE

avalonche:
why is there a need to forward to the http api? Is it because the http API could be gated by a different JWT token different from the auth endpoint?
In that case rollup-boost and optionally spin another endpoint to serve http endpoint requests

0xOsiris:
[...]

PR #61, opened 2025-01-27
feat(da-throttling): Forward DA throttling and other `miner_` requests to the builder 

0xKitsune:
ref #48 

This PR introduces logic to support DA throttling by forwarding `miner_ ` requests to the builder. 

For `miner_setMaxDALimit`, [the  MinerApiExtServer from op-alloy-rpc-jsonrpsee](https://docs.rs/op-alloy-rpc-jsonrpsee/0.9.6/op_alloy_rpc_jsonrpsee/traits/trait.MinerApiExtServer.html) was implemented. Since the [MinerApi
from Reth is synchronous](https://github.com/paradigmxyz/reth/blob/e2c45f2b1fd5a0c50f2440175a61e23410ca06ee/crates/rpc/rpc/src/miner.rs#L14-L24) and the `rollup-boost` implementation has async calls, I recreated the trait within `rollup-boost` to be async. We can open an issue on Reth and use the trait from `reth-rpc-api` once this is upstreamed.

Feel free to let me know any feedback on the approach and if you have a preference on how to extend or amend the current test harness to best test this feature.

PR #69, opened 2025-01-31
feat(da-throttling): Introduce DA throttling via the `MinerAPI`

### LINKED ISSUE #48: DA throttling

### LINKED ISSUE #63: Bug: Only Engine API server modules are initialized in the Binary
The only RPC methods actually initialized in `rollup-boost` is from the Engine API server [here](https://github.com/flashbots/rollup-boost/blob/81a921666d58ce59bc7f6c41b3dec9f2bbb6e407/src/main.rs#L209). 

I noticed this while testing #61 on a local Devnet. 

On `main` this means `eth_sendRawTransaction` is not a registered RPC method on the Server.

0xKitsune:
This PR introduces DA throttling to `rollup-boost` as well as fixes RPC module initialization. Note that this PR builds on #61, incorporating its changes.

### Closes #48

This PR introduces DA throttling by implementing the [MinerApiExtServer](https://docs.rs/op-alloy-rpc-jsonrpsee/0.9.6/op_alloy_rpc_jsonrpsee/traits/trait.MinerApiExtServer.html) as an RPC module. Additionally, a `MinerApi` trait has been implemented for the remaining `miner_` RPC endpoints. Tests were added for all newly introduced RPC modules.

While Reth provides a [MinerApi](https://github.com/paradigmxyz/reth/blob/e2c45f2b1fd5a0c50f2440175a61e23410ca06ee/crates/rpc/rpc/src/miner.rs#L14-L24), it is synchronous where `rollup-boost` requires an async implementation. To address this we recreated the trait within `rollup-boost` and can upstream this to Reth, allowing us to switch to `reth-rpc-api` once an async version is available.

Since the `auth_rpc` does not expose the Miner API, we introduced a new `ExecutionClient` struct to orchestrate requests to either the authenticated Engine API (via `auth_client`) or the JSON-RPC API (via `client`).  

```rust
/// Client interface for interacting with an execution layer node.
///
/// - **Engine API** calls are faciliated via the `auth_client` (requires JWT authentication).
/// -  All other API calls including the **Eth & Miner APIs** are faciliated via the `client` (optional JWT authentication).
///
#[derive(Clone)]
pub struct ExecutionClient<C: ClientT = HttpClient<HttpBackend>> {
    /// Handles requests to Eth, Miner, and other execution layer APIs (optional JWT authentication)
    pub client: Arc<C>,
    /// Address of the RPC server for execution layer API calls, excluding the Engine API
    pub http_socket: SocketAddr,
    /// Handles requests to the authenticated Engine API (requires JWT authentication)
    pub auth_client: Arc<HttpClient<AuthClientService<HttpBackend>>>,
    /// Address of the RPC server for authenticated Engine API calls
    pub auth_socket: SocketAddr,
}
```

### Fixes #63

Previously,the only RPC methods initialized were from the Engine API. This has been updated to ensure that `EthApiServer`, `MinerApiServer`, and `MinerApiExtServer` are properly merged into the RPC module. 

```rust
// --snip--
fn try_into(self) -> Result<RpcModule<()>, Self::Error> {
    let mut module: RpcModule<()> = RpcModule::new(());
    module.merge(EngineApiServer::into_rpc(self.clone()))?;
    module.merge(EthApiServer::into_rpc(self.clone()))?;
    module.merge(MinerApiServer::into_rpc(self.clone()))?;
    module.merge(MinerApiExtServer::into_rpc(self))?;

    Ok(module)
}
```

### Addresses #65
Since we introduced a `client` and `auth_client` for each execution client, we made a small update to the proxy logic to ensure that all non-multiplexed methods are routed to the auth RPC of the default L2 client with the proper authorization header.


0xOsiris (approved):

src/proxy.rs
```diff
@@ -88,18 +94,24 @@ where
...
                 // let rpc server handle engine rpc requests
                 let res = inner.call(req).await.map_err(|e| e.into())?;
                 Ok(res)
             } else {
+                info!(target: "proxy::call", message = "forwarding request to l2 client", "method" = ?rpc_request.method);
                 // Modify the URI
-                *req.uri_mut() = target_url;
+                *req.uri_mut() = target_url.clone();
+                // Insert JWT Authorization headers
+                req.headers_mut()
```
avalonche, line 113: why is there a need to insert the auth header here when it can be cloned from the original op-node request?

avalonche (commented):
could you update the .env.example and README with the new cli commands?

src/proxy.rs
```diff
@@ -88,18 +94,24 @@ where
 
             debug!(
```
avalonche, line 95: this log can be remove since there is info logging of the rpc method

src/main.rs
```diff
@@ -164,91 +131,67 @@ async fn main() -> Result<()> {
...
+        l2_client_args.l2_timeout,
+    )?;
+
+    let builder_args = args.builder;
+    let builder_jwt = if let Some(secret) = builder_args.builder_auth_jwtsecret {
+        secret
+    } else if let Some(path) = builder_args.builder_auth_jwtsecret_path.as_ref() {
+        JwtSecret::from_file(path)?
+    } else {
+        bail!("Missing Builder JWT secret");
```
avalonche, line 159: this would make the builder_jwt mandatory. Could it fallback to the l2 jwt token?

src/server.rs
```diff
@@ -356,7 +550,7 @@ where
                 }
             };
             if payload_status.is_invalid() {
-                error!(message = "builder payload was not valid", "url" = builder.url, "payload_status" = %payload_status.status, "payload_id" = %payload_id);
+                error!(message = "builder payload was not valid",  "payload_status" = %payload_status.status, "payload_id" = %payload_id);
```
avalonche, line 553: missing ?self.builder_client.auth_socket?

src/proxy.rs
0xOsiris: Because `op-node` is not the only service pointing at rollup-boost. `op-batcher` for example will be sending `eth_getBlockNumber` to the `http.api` of `op-geth`, but we are pointed at the Auth API, therefore we need to override the headers.

0xKitsune:
I have added a set of `FORWARD_REQUESTS` and updated the `ProxyService` to implement the forwarding logic. 

```rust
  fn call(&mut self, req: HttpRequest<HttpBody>) -> Self::Future {
     // --snip--

            if MULTIPLEX_METHODS.iter().any(|&m| method.starts_with(m)) {
                if FORWARD_REQUESTS.iter().any(|&m| method.starts_with(m)) {
                    let builder_client = client.clone();
                    let builder_req =
                        HttpRequest::from_parts(parts.clone(), HttpBody::from(body_bytes.clone()));
                    let builder_method = method.clone();

                    tokio::spawn(async move {
                        let _ = forward_request(
                            builder_client,
                            builder_req,
                            &builder_method,
                            builder_uri,
                            builder_auth,
                        )
                        .await;
                    });

                    let l2_req = HttpRequest::from_parts(parts, HttpBody::from(body_bytes));
                    info!(target: "proxy::call", message = "proxying request to rollup-boost server", ?method);
                    forward_request(client, l2_req, &method, l2_uri, None).await

       // --snip--
    }

```

I also removed the `EthApi`, `MinerApi` and `MinerApiExt` impls from `server.rs` in favor of the `forward_request` logic in the proxy. I will update tests, address all remaining comments and clean up the PR tomorrow for another round of review.

src/proxy.rs
```diff
@@ -74,48 +106,91 @@ where
...
+                            &builder_method,
+                            builder_uri,
+                            builder_auth,
+                        )
+                        .await;
+                    });
+
+                    let l2_req = HttpRequest::from_parts(parts, HttpBody::from(body_bytes));
+                    info!(target: "proxy::call", message = "proxying request to rollup-boost server", ?method);
+                    forward_request(client, l2_req, &method, l2_uri, None).await
```
0xOsiris, line 138: @0xKitsune Why are we forwarding to the L2 clients auth URI, and the builder RPC URI. This will fail to send all `miner_` requests to the L2 client currently as the miner API doesn't exist on the Auth API

0xOsiris: I think we need to pass both the l2 Auth URI, and the l2 RPC URI into the proxy layer. 

`FORWARD_REQUESTS` can target the RPC URI. 

Requests not in `MULTIPLEX_METHODS` can target the Auth URI.

src/main.rs
0xOsiris: I don't have a strong opinion here, but it seems a more Robust CLI configuration is to enforce the Builder, and L2 Client JWT are provided explicitly even if they are the same. In most configuration's it seems very likely they will not be the same, so silently inferring they are identical could cause silent bugs after startup. 

Let me know what you think

ferranbt (approved):

ferranbt (approved):

PR #78, opened 2025-02-09
Do not query builder if it is not building the block

ferranbt:
With this PR we do not query get_payload from block builder if we know that it is not building the block during no_tx_pool request calls. Before, this showed as errors in the logs as the builder did not have any reference to the payload.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -375,6 +375,19 @@ impl RollupBoostServer {
                 };
             });
         } else {
+            // If no payload attributes are provided, the builder will not build a block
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -410,6 +424,23 @@ impl RollupBoostServer {
                 .await
                 .unwrap_or(payload_id);
```
ferranbt, line 425: It could be simpler if we removed this and just stop if the payload_id is not on the context.
```suggestion
                .unwrap_or(payload_id);
```

PR #95, opened 2025-02-15
Cleanup Integration + builder down integration test

ferranbt:
This PR does several things:
- It cleans up a little bit of the integration framework (introduces functions to start/stop/restart the services).
- It connects through P2P the EL and Builder instances of the integration test harness.
- It introduces a new test that checks that the builder can be stopped, rollup-boost does not fail and once the builder gets restarted the system goes back to normal.


src/integration/mod.rs
```diff
@@ -128,17 +130,21 @@ impl ServiceCommand {
...
+pub struct ReadyParams {
+    pub log_pattern: String,
+    pub duration: Duration,
+}
+
 pub trait Service {
     fn command(&self) -> ServiceCommand;
-    fn ready(
-        &self,
-        service_instance: &mut ServiceInstance,
```
ferranbt, line 135: Changed this because it was really complex to model in Rust.

PR #96, opened 2025-02-17
Add a proxy in between Rollup-boost + Rbuilder and simulate latencies

ferranbt:
This PR adds a proxy in between Rollup-boost and the builder and adds two tests:
- One that simulates latency between Rollup-boost and the builder.
- One that changes the state root from the payload returned by the builder. Rollup-boost has to detect it and return the L2 block.

### DIFF
--- a/src/integration/proxy.rs
+++ b/src/integration/proxy.rs
@@ -1,4 +1,5 @@
 use bytes::Bytes;
+use http::header;
 use http_body_util::{combinators::BoxBody, BodyExt, Full};
 use hyper::client::conn::http1::Builder;
 use hyper::server::conn::http1;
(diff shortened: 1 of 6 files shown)


src/integration/proxy.rs
```diff
@@ -0,0 +1,141 @@
...
+    error: Option<JsonRpcError>,
+    id: Value,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct JsonRpcError {
+    code: i32,
+    message: String,
+    data: Option<Value>,
+}
```
avalonche, line 38: don't these types already exist in the rpc library?

avalonche (approved):

src/integration/proxy.rs
ferranbt: I tried to use them but they had some generics that I did not manage to figure out. Feel free to try.

Issue #91, opened 2025-02-12
Circuit Breaker


dmarzzz:
Similar to the idea of a circuit breaker for mev-boost on L1, rollup boost should auto shut off a block builder if multiple invalid payloads come from that block builder. TBD what turning the builder back on looks like. https://hackmd.io/@ralexstokes/BJn9N6Thc

ferranbt:
I am on the fence about this change since we are already validating with the fallback EL node each payload from the builder. So, the main thing we are skipping is the get payload call in case the builder is faulty. 

My assumption (and maybe we should document this more) is that the operator should track the error messages when the validation fails and manually set the builder in dry-run mode in the meantime.

This is a nice thing to have, but, not sure if worth it given that we want to keep RB as simple as possible.

dmarzzz:
If we expect an operator to do this themselves then I think it's pretty fair to put it inside the code, and relatively small diff. I hear you on simplicity though. If initial chains don't expect to use this then I'd say let's pause it for now. Will reach out (can't @ on gh mobile right now for some reason)

avalonche:
The main difference between this and l1 is that the relay is doing the block validation in which the circuit breaker design made more sense. Agree that since the sequencer is doing its own validation there is less value as the faulty builder payload gets discarded regardless

dmarzzz:
okay let's leave it open until my RFC from rollup boost operators gets answered, but will default assume we're not merging this feature.

dmarzzz:
@trianglesphere said let's kill it IRL and no response from anyone else so let's cancel this feature sorry @avalonche

PR #103, opened 2025-02-19
Add circuit breaker

avalonche:
Resolves https://github.com/flashbots/rollup-boost/issues/91

Notes:
- Only stops all engine calls to the builder (not forwarded calls - can add as a follow up if needed)
- Default to 5 slots 
- Updated README
- Adds test for circuit breaker condition


src/server.rs
```diff
@@ -285,7 +297,12 @@ impl RollupBoostServer {
 
         // TODO: Use _is_block_building_call to log the correct message during the async call to builder
         let (should_send_to_builder, _is_block_building_call) =
-            if let Some(attr) = payload_attributes.as_ref() {
+            if *self.num_invalid_builder_payloads.lock().await
+                >= self.config.max_allowed_invalid_builder_payloads
+            {
+                // Circuit breaker is triggered, do not send to builder
+                (false, false)
```
avalonche, line 304: @ferranbt for simplicity I set `_is_block_building_call` to false here. what uses did you have in mind for this variable? Is it fine to set to false when the circuit breaker is hit?

PR #106, opened 2025-02-19
Add wss support

cody-wang-cb:
Add native-tls to `tokio-tungstenite` to support wss


avalonche (approved):

ferranbt:
Merging since the lint errors come from the target branch which is not main.

Issue #107, opened 2025-02-20
Debug API dry run status


avalonche:
Json rpc api to return whether dry run mode is enabled

Issue #82, opened 2025-02-10
Dry Run Mode


dmarzzz:
**Is your feature request related to a problem? Please describe.**

Currently, Rollup Boost lacks a way to test interactions with external block builders in a production environment without jeopardizing the OP stack liveness. This makes network upgrades an all-or-nothing scenario, which introduces risk.

**Describe the solution you'd like**

Introduce a `--dry-run` flag to Rollup Boost that allows operators to:

- Propagate block requests to external builders without delivering them to the OP stack sequencer.

- Log interactions and outcomes for analysis.

- Expose metrics to verify the dry run mode's functionality.

**Implementation Details:**

- **Dry Run Flag:** Add a command-line option `--dry-run` that, when enabled, processes block requests from external builders but prevents their delivery to the OP stack sequencer.

- **Logging:** logging to indicate when Rollup Boost is operating in dry run mode, detailing interactions with external builders and any deviations from expected behavior.

**Exposed Metrics:**

To ensure confidence in the dry run mode's functionality, the following metrics should be exposed:

1. **Total Block Requests Sent:** The cumulative number of block requests initiated to external builders during the dry run.

2. **Successful Block Responses:** The count of valid block responses received from external builders.

3. **Failed Block Responses:** The number of block responses that failed validation or were otherwise deemed invalid.

4. **Average Response Time:** The mean time taken to receive responses from external builders after initiating a request.

5. **Skipped Block Deliveries:** The count of blocks that were processed but intentionally not delivered to the OP stack sequencer due to dry run mode.

6. **Error Logs:** Detailed logs of any errors encountered during interactions with external builders in dry run mode.

**Describe alternatives you've considered**

Currently, the only alternative is to test interactions in a separate staging environment, which may not fully replicate production conditions.

paranoiacblack:
Can we add some JSON-RPC commands that also allow us to control the mode that `rollup-boost` runs in (either disabled, dry-run or enabled)? For example, on our sequencers, we have admin commands that allow us to disable/enable the active sequencer and other such things: https://docs.optimism.io/operators/node-operators/json-rpc#admin. This is preferred to having to restart the node to run rollup-boost in a different mode.

PR #109, opened 2025-02-20
Add execution mode

ferranbt:
This PR extends the concept of the dry-run mode introduced in #82 into an execution mode with three different states:
- enabled: The builder receives all the engine API calls from RB.
- dry-run: The builder receives all the engine API calls from RB except for the payload request.
- disabled: The builder does not receive any API calls from RB.

I have modified the previous command:
```
$ rollup-boost debug dry-run
```
 
into a more generalized version:
```
$ rollup-boost debug set-execution-mode [enabled|dry-run|disabled]
```

and 

```
$ rollup-boost debug execution-mode
Execution-mode: <>
```

I updated the dry-run integration test to cover also the disabled test.

It will be up to a followup PR to decide whether we want to disable the multiplexed methods as well (i.e. miner).


ferranbt:
Closes #107

src/server.rs
```diff
@@ -103,14 +104,35 @@ impl PayloadTraceContext {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, clap::ValueEnum)]
+pub enum ExecutionMode {
```
avalonche, line 108: can you add serde rename to snake case

src/debug_api.rs
```diff
@@ -112,11 +128,12 @@ mod tests {
...
     async fn test_debug_client() {
         // spawn the server and try to modify it with the client
-        let dry_run = Arc::new(Mutex::new(false));
+        let execution_mode = Arc::new(Mutex::new(ExecutionMode::Enabled));
 
-        let server = DebugServer::new(dry_run.clone());
+        let server = DebugServer::new(execution_mode.clone());
         let _ = server.run(None).await.unwrap();
 
+        /*
```
avalonche, line 136: can the commented code be removed?

avalonche (approved):

PR #111, opened 2025-02-20
Add chaos testing

ferranbt:
This is still a draft since the setup is not that good yet.

The idea is that we can use a setup to deploy an Op-stack network (with Rollup-boost) and then create a unit test to deploy a reth builder with a proxy in front using the same utilities we use for integration tests.

Once that is done, we can randomly add delays and drop connections from the proxy to simulate chaos. Our success criteria is that even in this scenario, the liveness of the chain is not affected (there is a block every n seconds).

To deploy this setup you have to:

Run [this branch](https://github.com/ferranbt/optimism/tree/fb-v1.10.0) for the legacy op-stack docker-compose setup:
```
$ make devnet-up
```

Run `rollup-boost`:
```
$ cargo run -- --l2-url http://localhost:8551 --builder-url http://localhost:4444 --l2-jwt-token 688f5d737bad920bdfb2fc2f488d6b6209eebda1dae949a8de91398d932c517a --builder-jwt-token 688f5d737bad920bdfb2fc2f488d6b6209eebda1dae949a8de91398d932c517a
```

Run the unit test:
```
$ cargo test --features integration -- integration::integration_test::tests::test_integration_chaos --show-output --nocapture
```

Note that in the current unit test the genesis path is hardcoded to my computer. You would have to change that to test this.

My current thoughts:
- We should not use the docker-compose setup since it is a legacy system though I am not a fan of Kurtosis.
- The setup gets (way) more complicated if we have first to create a local Rollup-boost docker image and then run Kurtosis.

### DIFF
--- a/src/integration/integration_test.rs
+++ b/src/integration/integration_test.rs
@@ -6,7 +6,7 @@ mod tests {
     use std::time::Duration;
 
     use crate::debug_api::SetDryRunRequestAction;
-    use crate::integration::RollupBoostTestHarnessBuilder;
+    use crate::integration::{ChaosTestHarness, RollupBoostTestHarnessBuilder};
     use op_alloy_rpc_types_engine::OpExecutionPayloadEnvelopeV3;
 
     #[tokio::test]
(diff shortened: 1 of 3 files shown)


0xKitsune:
Closing for now due to inactivity. If this should be revisited, feel free to open a new issue/PR.

Issue #101, opened 2025-02-18
Record the backend with the latency of remote calls


trianglesphere:
Right now the latency metrics are over how long it takes rollup boost to process each operation. It would be more interesting to know how long each backend takes to process each engine API operation.

We can see how long rollup boost takes via metrics on op-node.

avalonche:
Are you referring to something as described in https://github.com/flashbots/rollup-boost/issues/67?

trianglesphere:
Yea, this looks like a dup of #67. I wasn't aware of that ticket and assume it was closed because there is the existing set of metrics.

ferranbt:
I think this issue covers a different topic than #67. We do not track how much time each call to L2 or the builder is taking from Rollup-boost.

Issue #100, opened 2025-02-18
Make FCU vs forkChoiceUpdated consistent in metrics


trianglesphere:
The count uses FCU, the latency uses fork_choice_updated

avalonche:
the two metrics can be merged, the latency metric is a histogram that already includes the count

PR #119, opened 2025-02-21
Metrics for engine client and proxy

avalonche:
- Refactor of the EngineClient to record metrics from a specific source
- Adds metrics for json rpc responses for all calls (do we need this? may add overhead for deserialising the json rpc response)
- resolves https://github.com/flashbots/rollup-boost/issues/101 https://github.com/flashbots/rollup-boost/issues/100


src/client.rs
```diff
@@ -38,6 +52,8 @@ impl ExecutionClient {
         auth_rpc: Uri,
         auth_rpc_jwt_secret: JwtSecret,
         timeout: u64,
+        metrics: Option<Arc<ServerMetrics>>,
+        payload_source: PayloadSource,
```
ferranbt, line 56: What is this one for?

src/metrics.rs
```diff
@@ -1,30 +1,172 @@
...
+    }
+
+    pub fn record_fork_choice_updated_v3(
+        &self,
+        latency: Duration,
+        code: String,
+        source: PayloadSource,
+    ) {
+        match source {
+            PayloadSource::L2 => {
```
ferranbt, line 123: This kills me a little bit.

src/metrics.rs
```diff
@@ -1,30 +1,137 @@
...
 
+use crate::server::PayloadSource;
+
 #[derive(Metrics)]
 #[metrics(scope = "rpc")]
 pub struct ServerMetrics {
-    #[metric(describe = "Count of forkchoice_updated_v3 calls proxied to the builder")]
-    pub fcu_count: Counter,
+    #[metric(describe = "Total latency for server `engine_newPayloadV3` call")]
+    pub new_payload_v3_total: Histogram,
```
ferranbt, line 12:
```suggestion
    pub new_payload_v3: Histogram,
```
I prefer the name without total prefix

src/metrics.rs
```diff
@@ -1,30 +1,137 @@
...
+    }
+}
+
+impl ClientMetrics {
+    pub fn new(source: &PayloadSource) -> Self {
+        Self {
+            new_payload_v3: histogram!("rpc.new_payload_v3", "target" => source.to_string()),
+            get_payload_v3: histogram!("rpc.get_payload_v3", "target" => source.to_string()),
+            fork_choice_updated_v3: histogram!("rpc.fork_choice_updated_v3", "target" => source.to_string()),
+            new_payload_v3_response_count: counter!("rpc.new_payload_v3_response_count"),
```
ferranbt, line 115: Doesnt histogram also cover counts?

ferranbt (approved):

src/metrics.rs
dmarzzz: Ideally there should be some indication of the metric's use in the name to help distinguish

src/metrics.rs
avalonche: its a bit hard to add the status code attribute without a separate metric since the target would be overwritten so I opted for a separate count metric

src/metrics.rs
avalonche: I think we can have a separate namespace between the client and total latencies, but its not as obvious as adding a total latency name to help distinguish the different metric

PR #120, opened 2025-02-22
Better LRU and Fix for Deadlock

0xForerunner:
I discovered a deadlock in the rollup boost server. I've posted a fix here using a concurrent, lock free LRU cache. In general this should make the code here more readable and less prone to errors.

As a general note, I've noticed that the way mutex locks are being handled in this code base seems to be pretty fast and loose. There were several cases where locks were being held far longer than they need to be.


src/server.rs
```diff
@@ -1,106 +1,96 @@
...
+                }
+                None => {
+                    vec![payload_id]
+                }
+            });
     }
 
-    async fn retrieve_by_parent_hash(&self, parent_hash: &B256) -> Option<Vec<Arc<BoxedSpan>>> {
-        let mut block_hash_to_payload_ids = self.block_hash_to_payload_ids.lock().await;
-        let mut payload_id_to_span = self.payload_id_to_span.lock().await;
```
0xForerunner, line 68: Deadlock can occur here, as locks are acquired in opposite order as in `fn store` above.

dmarzzz:
Thanks for the call out and contribution! Would you be willing to open up an issue for any other lock behavior usage you see as problematic? An audit locks issue would also work

0xForerunner:
@dmarzzz Yeah no problem! I'll probably just push another commit to this branch with a general lock cleanup. Already fixed a few so may as well keep it going :)

0xForerunner:
Okay so I think I've cleaned up all the locks excluding those used in tests. Those should probably be cleaned up as some point as well, but let's leave that for another PR.

@dmarzzz lemme know if this looks good to you :)

src/server.rs
```diff
@@ -1,110 +1,100 @@
...
-const CACHE_SIZE: usize = 100;
+const CACHE_SIZE: u64 = 100;
 
 pub struct PayloadTraceContext {
-    tracer: Arc<BoxedTracer>,
-    block_hash_to_payload_ids: Arc<Mutex<LruCache<B256, Vec<PayloadId>>>>,
-    payload_id_to_span: Arc<Mutex<LruCache<PayloadId, Arc<BoxedSpan>>>>,
-    local_to_external_payload_ids: Arc<Mutex<LruCache<PayloadId, PayloadId>>>,
+    tracer: BoxedTracer,
+    block_hash_to_payload_ids: Cache<B256, Vec<PayloadId>>,
```
avalonche, line 32: curious how this Cache differs from LruCache?

avalonche (approved):
LGTM! Thank you for the PR, makes the trace logic much cleaner

src/server.rs
0xForerunner: It just lock free, which means we don't need to worry about locking ourselves. Helps avoid errors! Should be significantly faster as well, not that it really matters in this case haha.

0xForerunner:
@avalonche feel free to merge this in when you're ready!

avalonche:
could you fix the checks? then it should be good to go

0xForerunner:
@avalonche sorry that test broke when I merged in main. Should be ready to go now.

0xForerunner:
@avalonche just bumping this so it doesn't fall behind again!

avalonche (approved):

0xForerunner:
@avalonche apparently nightly fmt doesn't work for our CI. Ran with stable fmt. Should work now :p

PR #121, opened 2025-02-26
feat: v4 payload support

mkykadir:


src/server.rs
```diff
@@ -686,12 +902,47 @@ mod tests {
...
+                            },
+                            blob_gas_used: 0xc0000,
+                        excess_blob_gas: 0x580000,
+                        },
+                        block_value: U256::from(0),
+                        blobs_bundle: BlobsBundleV1{
+                            commitments: vec![],
+                            proofs: vec![],
+                            blobs: vec![],
+                        },
```
avalonche, line 937: instead of duplicating execution v3 payload, can turn into variable

src/server.rs
```diff
@@ -36,6 +38,65 @@ pub struct PayloadTraceContext {
     local_to_external_payload_ids: Arc<Mutex<LruCache<PayloadId, PayloadId>>>,
 }
 
+enum PayloadVersion {
+    V3,
+    V4(Vec<Bytes>),
```
avalonche, line 43: instead of the payload version encapsulating the execution requests for the v4 new payload request, why not have a structure that encapsulates all the args for the new_payload method as an enum

src/server.rs
mkykadir: update in 2663591b06403397ce6f6089ee37cdd2eb76e6bb

src/server.rs
mkykadir: can you elaborate on this?

src/client.rs
```diff
@@ -156,6 +185,43 @@ impl ExecutionClient {
         response
     }
 
+    pub async fn new_payload_v4(
+        &self,
+        payload: ExecutionPayloadV3,
```
0xOsiris, line 190: Shouldn't this be `OpExecutionPayloadV4`

https://github.com/alloy-rs/op-alloy/blob/f7b7765d190ce19d88ae47d6cd58450366519cd5/crates/rpc-types-engine/src/payload/v4.rs#L14-L22

https://github.com/paradigmxyz/reth/blob/a765af9e6b7cf1b0332bb1d132ed09530f055c42/crates/optimism/rpc/src/engine.rs#L82

src/server.rs
```diff
@@ -265,14 +362,42 @@ impl EngineApiServer for RollupBoostServer {
...
         let elapsed = start.elapsed();
         if let Some(metrics) = &self.metrics {
             metrics.new_payload_v3_total.record(elapsed);
         }
         res
     }
+
+    async fn new_payload_v4(
+        &self,
+        payload: ExecutionPayloadV3,
```
0xOsiris, line 381: Same here - should be `OpExecutionPayloadV4` I believe

src/client.rs
0xOsiris: The v4 Engine API to my understanding is introduced in the `isthmus` hard fork adding an additional `withdrawalsRoot` to the execution payload

src/server.rs
mkykadir: `OpExecutionenvelopeV(3/4)` is from Alloy and fields are not similar to calls of `new_payload` but followed another approach in 4eb5e5187722ecaa7c7e43bff7e74f3eaa429738

src/client.rs
mkykadir: added in 16e9a4bc209be397198e98e7502f8ce58f002fbc

src/server.rs
mkykadir: added in 16e9a4bc209be397198e98e7502f8ce58f002fbc

PR #124, opened 2025-03-02
chore: allow configuration of debug server listen address

danyalprout:
### Description
Allow the configuration of the address which the debug server listens on. There are certain runtimes (e.g. docker compose), where each container is in its own networking space and to be able to communicate between the containers requires services to listen on `0.0.0.0` instead of `127.0.0.1`.

I've kept the default as `127.0.0.1` as it works on k8s and is the most restrictive.


avalonche (approved):
It'll be great if you could update .env.example and relevant sections in the README

PR #129, opened 2025-03-07
Add vercel support for book

sukoneck:
subj.


.github/workflows/book.yml
avalonche, line 1: was this meant to be deleted?

sukoneck: Yes, renamed to `book-tests` because it's just tests now. No strong feelings about file name.

PR #131, opened 2025-03-10
feat(proxy): add optional forward-url for l2 client and builder

0xOsiris:
Adds the following optional CLI arguments to `BuilderArgs` and `L2ClientArgs`
- `forward-url`
- `forward-jwt-token`
- `forward-jwt-path`

If left unspecified the `forward-url` for the builder, and the l2 client will default to the engine-API RPC endpoint making this change backwards compatible. 

Rationale:
Depending on the use case forwarded requests may be better suited to hit the HTTP API of the L2 Client, and the builder. 

For Example:
- `eth_sendRawTransactionConditional` only exists on the EthAPI, and not the engine-API on `op-geth` making rollup-boost incompatible with forwarding conditional 4337 transactions to the builder, and the sequencer.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -6,7 +6,7 @@ use alloy_rpc_types_engine::{
     ExecutionPayload, ExecutionPayloadV3, ForkchoiceState, ForkchoiceUpdated, JwtError, JwtSecret,
     PayloadId, PayloadStatus,
 };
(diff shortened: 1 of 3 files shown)


0xOsiris:
Closing in favor of using a dedicated transaction proxy service to serve this use case.

Issue #133, opened 2025-03-13
rollup-boost to choose block with higher gas used


avalonche:
Instead of default choosing the builder block, use whichever block used higher amount of gas. cc @trianglesphere

trianglesphere:
If both blocks are similar is fullness, we should probably use the remote builder. Maybe something like if the local block is 30% more full than the remote block we can use it.

The most invasive way to measure is to throw out all of the deposits + the builder system txn and then measure only gas of the user transactions, but even then, we'd only want to use the local block if it is substantially more full.

avalonche:
Happy with this approach, we can have this value configurable and tune it based on load testing scenarios.

dmarzzz:
Another valid heuristic could be if there is more than the builder txn + deposit txns then include. not sure i'd argue for it but it would also satisfy our purposes

trianglesphere:
Yea, number of transactions (probably with a similar percentage error) could be decent. The only case I worry about is if we end up with different transaction composition (i.e. lots of small vs a single large one), the large one is still valid.

PR #134, opened 2025-03-14
Add config for block selection

avalonche:
#133


src/server.rs
```diff
@@ -509,7 +512,21 @@ impl RollupBoostServer {
...
-            (Ok(builder), _) => Ok(builder),
+            (Ok(builder), Ok(l2)) => {
+                let builder_gas_used = builder
+                    .0
+                    .execution_payload
+                    .payload_inner
+                    .payload_inner
+                    .gas_used;
+                let l2_gas_used = l2.0.execution_payload.payload_inner.payload_inner.gas_used;
+                if builder_gas_used >= l2_gas_used * self.l2_boost_factor {
```
0xForerunner, line 523: you're multiplying by an integer here, not a float, so I think we have incorrect behaviour.

src/main.rs
```diff
@@ -100,6 +100,10 @@ struct Args {
     /// Execution mode to start rollup boost with
     #[arg(long, env, default_value = "enabled")]
     execution_mode: ExecutionMode,
+
+    /// percentage multipler to the local payload gas used for choosing payloads
+    #[arg(long, env, default_value = "0")]
+    l2_boost_factor: u64,
```
0xForerunner, line 106: I'm not sure l2_boost_factor is a very descriptive name. Maybe we can go with something like `--builder-gas-requirement=0.95` and parse a float here instead?

src/server.rs
```diff
@@ -133,6 +134,7 @@ impl RollupBoostServer {
         boost_sync: bool,
         metrics: Option<Arc<ServerMetrics>>,
         initial_execution_mode: ExecutionMode,
+        l2_boost_factor: u64,
```
0xForerunner, line 137: We probably want some input validation.

0xForerunner (changes requested):

src/server.rs
avalonche: I'm thinking you can specify more than 100 if you want the builder block to have significantly higher gas used than the local block

avalonche:
@0xForerunner I updated with multiple configs if you can do a re-review

PR #135, opened 2025-03-14
Tracing Refactor and Cleanup

0xForerunner:
This PR contains a handful of improvements:
- Bump dependency version
- Use `tracing-opentelemetry` crate
- Clean up logging
- Improve client error handling
- Automate span based metrics


src/tracing.rs
```diff
@@ -0,0 +1,156 @@
...
+        let provider = provider_builder.build();
+        let tracer = provider.tracer(env!("CARGO_PKG_NAME"));
+
+        let trace_filter = Targets::new()
+            .with_default(LevelFilter::OFF)
+            .with_target(&filter_name, LevelFilter::TRACE);
+
+        let registry = registry.with(OpenTelemetryLayer::new(tracer).with_filter(trace_filter));
+
+        match args.log_format {
```
0xOsiris, line 110: I would just consolidate this into a utility function to reduce the code duplication below

src/client.rs
```diff
@@ -1,49 +1,61 @@
...
     #[error(transparent)]
     Jwt(#[from] JwtError),
 }
 
+impl From<ClientError> for ErrorObjectOwned {
+    fn from(err: ClientError) -> Self {
+        match err {
+            ClientError::Jsonrpsee(err) => match err {
+                jsonrpsee::core::ClientError::Call(error_object) => error_object,
+                e => ErrorObjectOwned::owned(0, e.to_string(), Option::<()>::None),
```
0xOsiris, line 40: Is defaulting to `0` an appropriate status code here?

src/server.rs
```diff
@@ -302,301 +258,168 @@ impl RollupBoostServer {
...
+        fields(
+            otel.kind = ?SpanKind::Server,
+            %payload_id
+        )
+    )]
     async fn get_payload_v3(
         &self,
         payload_id: PayloadId,
     ) -> RpcResult<OpExecutionPayloadEnvelopeV3> {
-        info!(message = "received get_payload_v3", "payload_id" = %payload_id);
```
0xOsiris, line 421: Probably should retain the same verbosity on the trace

src/client.rs
0xForerunner: Ah yeah thanks for bringing this up. I believe these error codes are impl specific so I think it's fine. However could use some more thought here. Either way it's probably better than just submitting InternalError all the time.

src/tracing.rs
0xForerunner: Because adding layers changes the underlying types this doesn't work. There might be some weird trick you can do with type erasure haha but I don't think that's worth it.

I left a comment above related to the weird control flow for exactly this reason.

src/client.rs
0xForerunner: I take it back 0 is not the right choice. Will switch to 13.

0xForerunner:
@avalonche @ferranbt Could I grab a review?

src/client.rs
```diff
@@ -1,49 +1,61 @@
...
 }
 
+impl From<ClientError> for ErrorObjectOwned {
+    fn from(err: ClientError) -> Self {
+        match err {
+            ClientError::Jsonrpsee(jsonrpsee::core::ClientError::Call(error_object)) => {
+                error_object
+            }
+            // Status code 13 == internal error
+            e => ErrorObjectOwned::owned(13, e.to_string(), Option::<()>::None),
```
avalonche, line 42: is there a way to map different codes for different ClientErrors? also can probably extract the code into a const

0xForerunner: Yeah certainly we can do that. The above impl is similar to how it was done before, so perhaps that improvement can be left for another PR?

Let's create an issue for it and spec out some error codes.

src/client.rs
```diff
@@ -52,115 +64,105 @@ impl ExecutionClient {
...
         Ok(Self {
-            auth_client: Arc::new(auth_client),
+            auth_client,
             auth_rpc,
-            metrics,
             payload_source,
         })
     }
 
+    #[instrument(
```
avalonche, line 82: how does this record the status code and payload source in the response for metrics? I see a lot of the metrics code have been removed

src/integration/integration_test.rs
```diff
@@ -273,15 +273,6 @@ mod tests {
             assert!(block_creator.is_l2(), "Block creator should be the builder");
         }
 
-        // check that at some point we had the log "builder payload was not valid" which signals
-        // that the builder returned a payload that was not valid and rollup-boost did not process it.
-        let rb_service = harness._framework.get_mut_service("rollup-boost")?;
```
avalonche, line 278: assertion removed here? can you assert that an invalid payload error was returned in the integration test

src/main.rs
```diff
@@ -300,29 +263,6 @@ async fn main() -> eyre::Result<()> {
...
-        Ok(provider) => {
-            let _ = global::set_tracer_provider(provider);
-        }
-        Err(e) => {
-            error!(message = "failed to initiate tracing provider", "error" = %e);
-        }
-    }
-}
-
 async fn init_metrics_server(addr: SocketAddr, handle: PrometheusHandle) -> eyre::Result<()> {
```
avalonche, line 266: can move this function to the metrics file?

src/client.rs
0xForerunner: @avalonche metrics for the duration of each span will be created automatically using the span processor. If an instrumented function named 'foo' is called, we will get a histogram metric created with name 'foo_duration'. Histograms also contain a 'count' so we get all the same info we had before, but much cleaner.

src/tracing.rs
```diff
@@ -0,0 +1,156 @@
...
+    fn on_start(&self, _span: &mut opentelemetry_sdk::trace::Span, _cx: &opentelemetry::Context) {}
+
+    fn on_end(&self, span: opentelemetry_sdk::trace::SpanData) {
+        let duration = span
+            .end_time
+            .duration_since(span.start_time)
+            .unwrap_or_default();
+
+        // Remove status description to avoid cardinality explosion
+        let status = match span.status {
```
avalonche, line 30: how is the status set in the span? can you have the error code as the label instead? That way there will be specificity for the error without a cardinality explosion

src/integration/integration_test.rs
0xForerunner: Assertion was removed here because the log is no longer created. Let me see if I can find a work around to add this back. For the future, I'm not sure log parsing is the best way to verify integration tests, though off the top of my head I don't have a better solution :p I'll give it a think.

src/main.rs
0xForerunner: This is done in one of my other PRs already :)

src/tracing.rs
avalonche: the instrument macro seem to only have payload id, payload source etc as fields

0xForerunner: We can definitely add the error code to the labels! Good idea!

src/server.rs
```diff
@@ -302,301 +258,168 @@ impl RollupBoostServer {
...
-                message = "returning block",
-                "hash" = %block_hash,
-                "number" = %block_number,
-                "context" = %context,
-                "payload_id" = %payload_id
-            );
-            payload
-        })
+        let (payload, context) = match (builder_payload, l2_payload) {
+            (Ok(Some(builder)), _) => Ok((builder, "builder")),
```
avalonche, line 345: can replace string with PayloadSource enum here

0xForerunner: Ah nice good call, TY.

src/server.rs
```diff
@@ -302,301 +258,168 @@ impl RollupBoostServer {
...
-            (Err(_), Ok(l2)) => Ok(l2),
-            (Err(_), Err(e)) => Err(e),
-        };
-        payload.map(|(payload, context)| {
-            let inner_payload = ExecutionPayload::from(payload.clone().execution_payload);
-            let block_hash = inner_payload.block_hash();
-            let block_number = inner_payload.block_number();
-
-            if let Some(metrics) = &self.metrics {
-                metrics.increment_blocks_created(&context);
```
avalonche, line 522: note that we also want to record whether the server used the builder or local payload here (e.g. if the builder payload is invalid, was discarded due to some other block selection mechanism) which looks to be removed from the instrumentation

0xForerunner: This value can be derived from the client get_payload_v3 spans. You can just look at spans where span kind is client, payload source is builder, and status is ok. A histogram with count is automatically created for this.

src/tracing.rs
0xForerunner: @avalonche do you mind if we create an issue for this, or would you like this to be handled in this PR?

src/client.rs
0xForerunner: Oh I think I see what you're saying here. Yea that's correct, the response code is no longer included as a metric label. I just had a look about how we could achieve that in a nice way and came across [this issue](https://github.com/tokio-rs/tracing/issues/3190) in tracing. Without the ability to do that, keeping the response codes as labels would be pretty messy. I personally don't see a huge loss here since we still span status as a label, but if you feel strongly about it I'm sure I can implement a work around?

src/tracing.rs
0xForerunner: @avalonche I think I found a pretty ergonomic way to set error codes in the span/metric labels. Just pushed a new commit with the updates, let me know what you think.

src/integration/integration_test.rs
0xForerunner: Fixed in most recent commit

0xForerunner:
@avalonche I believe I've addressed all your comments :) Please review my most recent 2 commits and let me know if everything looks good.

src/server.rs
avalonche: the client get payload span only asserts that the response is ok, not necessarily what payload was chosen by rollup-boost. If the builder payload fails validation the metrics would not capture that the local payload was chosen

avalonche (approved):
Changes look ok to me but please add outstanding comments as issues on the repo

avalonche:
also would be great to update the docs on the changes to metrics. I believe there are breaking changes to the metric labels which would break dashboards

src/server.rs
```diff
@@ -302,301 +258,168 @@ impl RollupBoostServer {
...
-                            "url" = ?builder_client.auth_rpc,
-                            "error" = %e,
-                            "head_block_hash" = %fork_choice_state.head_block_hash
-                        );
-                    }
-                }
-                if let Some(mut s) = span {
-                    s.end()
-                };
+                let _ = builder_client
```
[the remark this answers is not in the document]
0xForerunner: Interesting. In practice they are the same at the very least.

src/client.rs
0xForerunner: #147

src/server.rs
0xForerunner: This line:
```rust
        tracing::Span::current().record("payload_source", context.to_string());
```

Sets the payload source, which is added as a label.

0xForerunner:
@avalonche I've updated the docs for you. Looks like I don't have permission to merge yet so you may have to do it for me! TY!

src/server.rs
avalonche: oh right, thanks for pointing that out! I think everything is at parity with the current implementation then

PR #143, opened 2025-03-21
feat(metrics): add metric for builder proxy being up

paranoiacblack:
Adds a metric, "rpc.builder_up" that indicates whether the builder proxy is responding to requests. If the builder proxy returns 500 Internal Server Error, the builder proxy is reported as down. Otherwise, the builder proxy is reported as up.

This is a valuable feature for chain operators running rollup boost who need to track the status of the remote builder.

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -1,6 +1,6 @@
 use std::time::Duration;
 
-use metrics::{Counter, Histogram, counter, histogram};
+use metrics::{Counter, Gauge, Histogram, counter, gauge, histogram};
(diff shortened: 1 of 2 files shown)


0xForerunner:
I'm not sure this is any more useful than just keeping a histogram for builder responses with a status label attached. I would personally prefer that approach. There's already an open PR which includes this functionality as well.

0xKitsune:
Closing for now as metrics to track builder responses were included in #135. Feel free to open an issue if there are additional metrics that should be added.

PR #149, opened 2025-03-27
Not using flashblock payload

cody-wang-cb:
When websocket is on for rollup-boost, there seems to be an issue with using the flashblock payload rather than calling `get_payload` whenever rollup-boost is restarted, it causes sequencing to stop for some reason. Root cause is still unknown but commenting it out for now.


ferranbt:
Trying to solve this as part of #157

ferranbt:
@cody-wang-cb Is this functionality still needed?

ferranbt:
Closing this since we added the flashblocks branch to main, please reopen there if the issue is still there

Issue #126, opened 2025-03-03
rollup-boost leaking TCP connections


avalonche:

dmarzzz:
<img width="1271" alt="Image" src="https://github.com/user-attachments/assets/23b410c1-abad-401f-b154-3992873cdca8" />

0xOsiris:
I am curious if this behavior has been observed  on the latest version of rollup-boost. I was stress testing rollup-boost yesterday, and analyzing the behavior of established connections over time. Throughout testing established connections grew, and shrank in relation to traffic, but went back to a consistent minimum almost immediately after traffic reduced to 0. 

I wasn't able to see any evidence of the RPC Client's holding onto idle connections

PR #155, opened 2025-03-28
Remove idle tcp connections

avalonche:
Resolves https://github.com/flashbots/rollup-boost/issues/126 for the proxy client. However there isn't an explicit way to configure the timeout on the json rpc see client library.

### DIFF
--- a/src/client/http.rs
+++ b/src/client/http.rs
@@ -1,3 +1,5 @@
+use std::time::Duration;
+
 use crate::server::PayloadSource;
 use alloy_rpc_types_engine::JwtSecret;
 use http::Uri;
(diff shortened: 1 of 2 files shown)


src/client/http.rs
```diff
@@ -32,7 +34,10 @@ impl HttpClient {
             .enable_http2()
             .build();
 
-        let client = Client::builder(TokioExecutor::new()).build(connector);
+        let client = Client::builder(TokioExecutor::new())
+            .pool_timer(hyper_util::rt::TokioTimer::new())
+            .pool_idle_timeout(Duration::from_millis(2_500))
```
0xOsiris, line 39: Let's make this a constant

0xForerunner:
@avalonche have you been able to actually verify that this fixes the problem?

PR #158, opened 2025-03-31
fix: forward `getBlockByHash` and `getBlockByNumber`

0x416e746f6e:
at `batcher`'s updates the sequencer will be hitting the builder with `eth_getBlockByHash` and `eth_getBlockByNumber` calls => we should proxy those to the remote builder as well.

### DIFF
--- a/src/proxy.rs
+++ b/src/proxy.rs
@@ -11,18 +11,22 @@ use std::{future::Future, pin::Pin};
 use tower::{Layer, Service};
 use tracing::info;
 
-const MULTIPLEX_METHODS: [&str; 4] = [
+const MULTIPLEX_METHODS: [&str; 6] = [
     "engine_",
-    "eth_sendRawTransactionConditional",
+    "eth_getBlockByHash",
+    "eth_getBlockByNumber",
     "eth_sendRawTransaction",
+    "eth_sendRawTransactionConditional",
     "miner_",
 ];
-const FORWARD_REQUESTS: [&str; 6] = [
+const FORWARD_REQUESTS: [&str; 8] = [
+    "eth_getBlockByHash",
+    "eth_getBlockByNumber",
     "eth_sendRawTransaction",
     "eth_sendRawTransactionConditional",
     "miner_setExtra",
-    "miner_setGasPrice",
     "miner_setGasLimit",
+    "miner_setGasPrice",
     "miner_setMaxDASize",
 ];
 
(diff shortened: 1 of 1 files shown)


0xOsiris:
> at `batcher`'s updates the sequencer will be hitting the builder with `eth_getBlockByHash` and `eth_getBlockByNumber` calls => we should proxy those to the remote builder as well.

The `op-batcher` only need's the return value from these requests from either the builder, or op-geth. This is a stateless request with respect to the builder - multiplexing these requests would not give the op-batcher any more information than it already has

0x416e746f6e:
> > at `batcher`'s updates the sequencer will be hitting the builder with `eth_getBlockByHash` and `eth_getBlockByNumber` calls => we should proxy those to the remote builder as well.
> 
> The `op-batcher` only need's the return value from these requests from either the builder, or op-geth. These are stateless requests with respect to the builder - multiplexing these requests would not give the op-batcher any more information than it already has

i.i.u.i.r., `op-batcher`'s job is to publish l2 updates back to l1's DA => it's completely out of picture after that happened (and this "after that happened" is exactly the scope of the present issue).

also, in current design batcher never interacts with `rollup-boost` (and consequently with the remote builder) directly.  but maybe it should..  thanks for flagging this @0xOsiris - I will need to think about it.

0xOsiris:
> > > at `batcher`'s updates the sequencer will be hitting the builder with `eth_getBlockByHash` and `eth_getBlockByNumber` calls => we should proxy those to the remote builder as well.
> > 
> > 
> > The `op-batcher` only need's the return value from these requests from either the builder, or op-geth. These are stateless requests with respect to the builder - multiplexing these requests would not give the op-batcher any more information than it already has
> 
> i.i.u.i.r., `op-batcher`'s job is to publish l2 updates back to l1's DA => it's completely out of picture after that happened (and this "after that happened" is exactly the scope of the present issue).
> 
> also, in current design batcher never interacts with `rollup-boost` (and consequently with the remote builder) directly. but maybe it should.. thanks for flagging this @0xOsiris - I will need to think about it.

The op-batcher need's to be pointed directly at rollup-boost as the l2 client. Otherwise `miner_setMaxDASize` request's will never hit the block builder in times when the pending queue of the batcher is too large to publish DA. In this case the builder will continue to build oversize payloads, and the health of the network is at risk to a spam attack.

Issue #156, opened 2025-03-28
Feat: Isthmus Hardfork Compatibility


0xOsiris:
Opening an issue to track v4 engine API integration into rollup-boost

PR #159, opened 2025-04-01
Add engine_api v4 support

ferranbt:
Main scaffolding is done, tests blocked by https://github.com/paradigmxyz/reth/issues/15435

## Builder-playground test

Build local docker image:
```
$ make docker-image
```

Clone builder-playground at run the op-stack with the new image:
```
$ go run main.go cook opstack --output output --external-builder op-reth --enable-latest-fork=0 --override rollup-boost=rollup-boost:latest
```

## Testing

We cannot do integration tests yet because there is not a nightly release available yet for `op-reth`. But, in the Kurtosis setup I am enabling the `Ishtmus` and `Prague` fork after 2 blocks.


ferranbt:
Closes #156

src/server.rs
```diff
@@ -272,6 +290,8 @@ impl EngineApiServer for RollupBoostServer {
             debug!(message = "execution mode is disabled, skipping FCU call to builder", "head_block_hash" = %fork_choice_state.head_block_hash);
         } else if should_send_to_builder {
             let builder_client = self.builder_client.clone();
+            println!("print payload attributes: {:?}", payload_attributes);
```
avalonche, line 293: debugging log?

0xOsiris:
Can we update the `kurtosis.yml` to enable Isthmus in CI, and also change the docker image of Reth to the nightly patch in kurtosis?

src/proxy.rs
```diff
@@ -132,7 +132,7 @@ where
                     });
 
                     let l2_req = HttpRequest::from_parts(parts, HttpBody::from(body_bytes));
-                    info!(target: "proxy::call", message = "proxying request to rollup-boost server", ?method);
+                    info!(target: "proxy::call", message = "forward request to rollup-boost server", ?method);
```
0xOsiris, line 135: I think it would make sense to prefix our targets with `rollup_boost` since our OTLP filter is on `rollup_boost` targets. 

https://github.com/flashbots/rollup-boost/blob/eca9266b99836b1dae525d7aebe93b7b6aeb3051/src/tracing.rs#L112

Not really relevant to this PR, but I think it would make sense to set `rollup_boost` as the target on all traces

src/integration/service_reth.rs
```diff
@@ -68,7 +68,8 @@ impl Service for RethConfig {
             })
             .arg("--color")
             .arg("never")
-            .arg("--ipcdisable");
+            .arg("--ipcdisable")
+            .arg("-vvvvv");
```
avalonche, line 72: do we want to enable tracing here on reth? seems it would be too verbose

src/integration/mod.rs
```diff
@@ -826,3 +849,25 @@ impl BlockBuilderCreatorValidator {
...
+    );
+
+    let deposit_tx = TxDeposit {
+        source_hash: B256::default(),
+        from: address!("DeaDDEaDDeAdDeAdDEAdDEaddeAddEAdDEAd0001"),
+        to: TxKind::Call(address!("4200000000000000000000000000000000000015")),
+        mint: None,
+        value: U256::default(),
+        gas_limit: 210000,
+        is_system_transaction: true,
```
avalonche, line 865: does this deposit tx work? had to set is_system_transaction: false for it to work in op-rbuilder

avalonche (approved):

src/integration/mod.rs
ferranbt: Our integration tests are running in an old (two weeks) version of Reth

src/proxy.rs
ferranbt: https://github.com/flashbots/rollup-boost/issues/163

0xOsiris (approved):
LGTM, nice work on this!

PR #165, opened 2025-04-04
Integration Test Improvements

0xForerunner:
This PR contains the following improvements:
- Removes local dependencies in favour of proper integration binaries and docker containers.
- Improves the reliability of some integration tests.
- Changes the integration test directory structure to be more standard.
- Removes the `integration` feature
- Parallelizes tests in CI with `cargo nextest`


src/client/rpc.rs
```diff
@@ -371,15 +371,28 @@ mod tests {
...
     use std::result::Result;
     use std::str::FromStr;
 
     use super::*;
 
-    const AUTH_PORT: u32 = 8550;
     const AUTH_ADDR: &str = "0.0.0.0";
     const SECRET: &str = "f79ae8046bc11c9927afe911db7143c51a806c4a537cc08e0d37140b0192f430";
 
+    fn get_available_port() -> u16 {
```
0xOsiris, line 383: Can't you just set the port to `0` and the OS will choose an available port?

src/lib.rs
```diff
@@ -1,8 +1,3 @@
-#![cfg_attr(
```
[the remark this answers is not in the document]
0xForerunner: Ah yeah got rid of that by accident, will add back.

tests/builder_returns_incorrect_block.rs
```diff
@@ -0,0 +1,69 @@
...
+        .proxy_handler(Arc::new(Handler))
+        .build()
+        .await?;
+
+    let mut block_generator = harness.block_generator().await?;
+
+    // create 3 blocks that are processed by the builder
+    for _ in 0..3 {
+        let (_block, block_creator) = block_generator.generate_block(false).await?;
+        assert!(block_creator.is_l2(), "Block creator should be the builder");
```
0xOsiris, line 57:
```suggestion
        assert!(block_creator.is_l2(), "Block creator should be the l2");
```

src/client/rpc.rs
```diff
@@ -371,15 +371,28 @@ mod tests {
...
+            let port: u16 = rand::random_range(1000..20000);
+            if port_is_available(port) {
+                return port;
+            }
+        }
+    }
+
+    fn port_is_available(port: u16) -> bool {
+        TcpListener::bind(("127.0.0.1", port)).is_ok()
+    }
```
0xKitsune, line 394: IIRC you should be able to do something like the following to get an available port

```rust
fn get_available_port() -> u16 {
    let listener = TcpListener::bind("127.0.0.1:0").expect("Failed to bind");
    listener.local_addr().unwrap().port()
}
```

0xForerunner: This is less reliable, since the OS will tend to choose ports that are next to each other. So if you call this function twice in a row you'll be more likely to get the same port each time.

0xForerunner: You could do this if you kept a global state of claimed ports.

0xForerunner: @0xKitsune I Added a ClaimedPorts hashset to prevent collisions from back to back calls.

0xForerunner:
So this with this last commit I've been able to fix all the flakey tests. I successfully rand 50 iterations of `cargo nextest run` without any failures. Should be ready to merge.

tests/common/services/op_reth.rs
```diff
@@ -0,0 +1,207 @@
...
+use serde_json::Value;
+use testcontainers::{
+    ContainerAsync, CopyToContainer, Image,
+    core::{ContainerPort, WaitFor},
+};
+
+use crate::common::TEST_DATA;
+
+const NAME: &str = "ghcr.io/paradigmxyz/op-reth";
+const TAG: &str = "v1.3.4";
```
0xOsiris, line 21:
```suggestion
const TAG: &str = "latest";
```

0xOsiris (approved):
LGTM, awesome work on this! This looks great

tests/common/services/op_reth.rs
0xForerunner: I think we probably want to intentionally increment the version, no? Otherwise we may have tests failing randomly.

0xOsiris: mmm, I guess I'd prefer to just always stay on latest or nightly, and address things as they come up

0xKitsune (approved):
Nice work!

PR #176, opened 2025-04-16
Reth 1.3.11

cody-wang-cb:
Support reth 1.3.11

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,9 +55,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "0.13.0"
+version = "0.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "27d301f5bcfd37e3aac727c360d8b50c33ddff9169ce0370198dedda36a9927d"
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -605,8 +649,18 @@ impl RollupBoostServer {
 
         let (l2_payload, builder_payload) = tokio::join!(l2_client_future, builder_client_future);
         let (payload, context) = match (builder_payload, l2_payload) {
-            (Ok(Some(builder)), _) => Ok((builder, PayloadSource::Builder)),
-            (_, Ok(l2)) => Ok((l2, PayloadSource::L2)),
+            (Ok(Some(builder)), Ok(l2)) => {
+                histogram!("block_building_gas_difference")
```
avalonche, line 653: these can be defined in `impl SpanProcessor for MetricsSpanProcessor {` on the end of the call?

src/client/rpc.rs
```diff
@@ -291,6 +302,11 @@ impl RpcClient {
             .set_code()?;
 
         if res.is_invalid() {
+            counter!(format!(
+                "rpc_new_payload_v4_invalid_payload_{}",
```
avalonche, line 306: many of the metrics can be simplified by having the method name and payload source as an attribute

avalonche (approved):

PR #177, opened 2025-04-16
loopback address for testing to bypass firewall issues

jowparks:
Note I did not change default values for `cli` as I didn't want to modify default behavior in fear of breaking some downstream default usage.


jowparks:
Closing in lieu of https://github.com/flashbots/rollup-boost/pull/176/files

PR #178, opened 2025-04-16
metrics updates

jowparks:
Adds a few metrics, also updates loopback IP in necessary places.

```
# TYPE rollup_boost_error counter
rollup_boost_error{code="-38001",span_kind="Client",status="error"} 1

# TYPE rollup_boost_block_building_gas_delta summary
rollup_boost_block_building_gas_delta{payload_source="builder",span_kind="Server",status="unset",quantile="0"} 0
...
rollup_boost_block_building_tx_count_delta_count{payload_source="builder",span_kind="Server",status="unset"} 1
```

### DIFF
--- a/src/client/rpc.rs
+++ b/src/client/rpc.rs
@@ -12,7 +12,6 @@ use http::Uri;
 use jsonrpsee::http_client::transport::HttpBackend;
 use jsonrpsee::http_client::{HttpClient, HttpClientBuilder};
 use jsonrpsee::types::ErrorObjectOwned;
-use metrics::counter;
 use op_alloy_rpc_types_engine::{
     OpExecutionPayloadEnvelopeV3, OpExecutionPayloadEnvelopeV4, OpExecutionPayloadV4,
     OpPayloadAttributes,
@@ -25,6 +24,9 @@ use thiserror::Error;
 use tracing::{error, info, instrument};
 
 const INTERNAL_ERROR: i32 = 13;
+const INVALID_PAYLOAD: i32 = 1337;
+const IO_ERROR: i32 = 1338;
+const JWT_ERROR: i32 = 1339;
 
 pub(crate) type ClientResult<T> = Result<T, RpcClientError>;
 
@@ -44,7 +46,8 @@ trait Code: Sized {
     fn code(&self) -> i32;
 
     fn set_code(self) -> Self {
-        tracing::Span::current().record("code", self.code());
+        let code_value: i64 = self.code().into();
+        tracing::Span::current().record("code", code_value.to_string());
         self
     }
 }
@@ -62,9 +65,10 @@ impl<T, E: Code> Code for Result<T, E> {
 impl Code for RpcClientError {
     fn code(&self) -> i32 {
         match self {
+            RpcClientError::InvalidPayload(_) => INVALID_PAYLOAD,
+            RpcClientError::Io(_) => IO_ERROR,
+            RpcClientError::Jwt(_) => JWT_ERROR,
             RpcClientError::Jsonrpsee(e) => e.code(),
-            // Status code 13 == internal error
-            _ => INTERNAL_ERROR,
         }
     }
 }
@@ -154,7 +158,7 @@ impl RpcClient {
         }
 
         if res.is_invalid() {
-            counter!("rpc_invalid_payload", "method" => "fork_choice_updated_v3", "payload_source" => self.payload_source.to_string()).increment(1);
+            error!("Invalid payload ({}): {:?}", self.payload_source, res);
             return Err(RpcClientError::InvalidPayload(
                 res.payload_status.status.to_string(),
             ))
@@ -172,6 +176,7 @@ impl RpcClient {
             target = self.payload_source.to_string(),
             url = %self.auth_rpc,
             %payload_id,
+            code,
         )
     )]
     pub async fn get_payload_v3(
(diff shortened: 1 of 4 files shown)


src/client/rpc.rs
```diff
@@ -153,6 +154,7 @@ impl RpcClient {
         }
 
         if res.is_invalid() {
+            counter!("rpc_invalid_payload", "method" => "fork_choice_updated_v3", "payload_source" => self.payload_source.to_string()).increment(1);
```
avalonche, line 157: https://github.com/flashbots/rollup-boost/pull/178/files#r2049347464

src/client/rpc.rs
```diff
@@ -213,6 +215,7 @@ impl RpcClient {
             .set_code()?;
 
         if res.is_invalid() {
+            counter!("rpc_invalid_payload", "method" => "new_payload_v3", "payload_source" => self.payload_source.to_string()).increment(1);
```
avalonche, line 218: https://github.com/flashbots/rollup-boost/pull/178/files#r2049347464

src/server.rs
```diff
@@ -605,6 +643,11 @@ impl RollupBoostServer {
 
         let (l2_payload, builder_payload) = tokio::join!(l2_client_future, builder_client_future);
         let (payload, context) = match (builder_payload, l2_payload) {
+            (Ok(Some(builder)), Ok(l2)) => {
```
avalonche, line 646: have you tested that the metrics show up? wondering if you need to add these metrics to the instrument macro
```
    #[instrument(
        skip_all,
        err,
        fields(
            otel.kind = ?SpanKind::Client,
            target = self.payload_source.to_string(),
            head_block_hash = %fork_choice_state.head_block_hash,
            url = %self.auth_rpc,
            code,
            payload_id
        )
    )]
```

src/server.rs
```diff
@@ -534,7 +568,11 @@ impl RollupBoostServer {
             let builder = self.builder_client.clone();
             let new_payload_clone = new_payload.clone();
             tokio::spawn(async move {
-                let _ = builder.new_payload(new_payload_clone).await;
+                let result = builder.new_payload(new_payload_clone).await;
+                if let Err(_) = &result {
+                    error!("Invalid payload (builder): {:?}", result);
+                    counter!("invalid_payload", "method" => "new_payload", "payload_source" => "builder").increment(1);
```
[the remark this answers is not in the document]
jowparks: Oh nice, good call, I will do that. Unfortunately it doesn't tell us where the invalid payload is being bubbled up from, but I think it is sufficient.

src/server.rs
jowparks: Great question, I think I missed this because I did not test. What is your workflow for testing metrics like this in a local dev environment?

src/tracing.rs
```diff
@@ -57,6 +57,34 @@ impl SpanProcessor for MetricsSpanProcessor {
             ])
             .collect::<Vec<_>>();
```
jowparks, line 59: Should this only be incremented on success code?

src/client/rpc.rs
```diff
@@ -61,8 +62,8 @@ impl<T, E: Code> Code for Result<T, E> {
 impl Code for RpcClientError {
     fn code(&self) -> i32 {
         match self {
+            RpcClientError::InvalidPayload(_) => INVALID_PAYLOAD,
```
avalonche, line 65: are there other rpc error types? also I wonder if the code is already in error for `RpcClientError::InvalidPayload(_)` instead of having to manually specify the code as a constant

src/server.rs
avalonche: we spin up a local op-stack with [builder-playground](https://github.com/flashbots/builder-playground) and then curl the metrics endpoint

src/tracing.rs
avalonche: what additional information does this metric provide that you can't already get from `rollup_boost_get_payload_v3_duration_count` filtering for payload_source and status = success?

src/client/rpc.rs
jowparks: The code will default to `INTERNAL_ERROR` I believe. I can add the other error types if you like (Io, jwt)

jowparks: I think the code span will be `13` currently for any of the spans with internal server error.

src/tracing.rs
jowparks: I guess it is just a rollup of the two types, so I will just sum over those two, thanks!

src/client/rpc.rs
avalonche: I meant the error inside `RpcClientError::InvalidPayload(e)` - I thought there were default JSON RPC error codes encoded

avalonche: otherwise is fine to me, can be added as a follow up

src/server.rs
jowparks: For posterity, the setup is similar to this: 
https://github.com/flashbots/rbuilder/blob/develop/crates/op-rbuilder/README.md#local-devnet

src/tracing.rs
```diff
@@ -57,6 +57,39 @@ impl SpanProcessor for MetricsSpanProcessor {
...
 
+        let code = span
+            .attributes
+            .iter()
+            .find(|attr| attr.key.as_str() == "code")
+            .and_then(|attr| attr.value.as_str().parse::<i32>().ok())
+            .unwrap_or(0);
+
+        if code != 0 {
+            counter!("error", &labels).increment(1);
```
avalonche, line 68: is there a need for the counter here? the code should be included in `histogram!(format!("{}_duration", span.name), &labels).record(duration);`

src/tracing.rs
```diff
@@ -57,6 +57,39 @@ impl SpanProcessor for MetricsSpanProcessor {
...
+        }
+
+        let gas_delta = span
```
avalonche, line 71: some documentation for this would be good i.e. positive number = builder used more gas / had more transactions. negative number = local block had more. A comment here will be fine

avalonche (approved):

src/tracing.rs
avalonche: no need to have the code here

avalonche (approved):

Issue #179, opened 2025-04-17
Fail Safe Switch to Auto Default to `op-geth` blocks in `get_payload`


0xOsiris:
Introduce an `ExecutionMode` variant `ExecutionMode::Fallback` that can be set over the debug API  with `debug_setExecutionMode`. The purpose of this mode of operation would be to be a failsafe switch that shuts off returning block builder payloads. 

When deploying rollup-boost to mainnet this gives a chain operator the ability to stop using the builder's payloads (while still sending it FCU's, and getPayload requests) to keep up with the tip which could be necessary if e.g. the builder is returning empty blocks, and transactions on the network are not being mined. 

Block selection policy is another approach to solve this issue, but this is not feasible on all chains. For example if the block builder has a stricter transaction validation policy than `op-geth` the happy path could have `op-geth` blocks with a higher gas consumption than builder blocks.

PR #180, opened 2025-04-17
feat: add fallback execution mode

0xOsiris:
Address #179

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -23,4 +23,5 @@ docs/
 
 # Scripts that aren't needed for runtime
 scripts/
+tests/
 **/integration
--- a/src/debug_api.rs
+++ b/src/debug_api.rs
@@ -152,5 +152,17 @@ mod tests {
         // Verify again with get_execution_mode
         let status = client.get_execution_mode().await.unwrap();
         assert_eq!(status.execution_mode, ExecutionMode::Enabled);
+
+        // Test setting fallback execution mode
+        let result = client
+            .set_execution_mode(ExecutionMode::Fallback)
+            .await
+            .unwrap();
+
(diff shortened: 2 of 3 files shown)


avalonche (approved):
Instead of adding an execution mode, wondering if it'll be more useful to have a generic trait that determines block selection between the builder vs local or potentially multiple builder payloads? which will be part of the api 
this approach is fine if you need something quick to implement a fallback

0xOsiris:
> Instead of adding an execution mode, wondering if it'll be more useful to have a generic trait that determines block selection between the builder vs local or potentially multiple builder payloads? which will be part of the api this approach is fine if you need something quick to implement a fallback

Yeah I agree here that this is also a solution (and definitely okay with adding this in as an option as well). The reason that solution doesn't work holistically for our use case is because World Chain's block builder has a stricter transaction validation policy than op-geth, so any block selection policy based on gas is not taking into account orphaned transactions in op-geth that failed transaction validation in the builder. So, there will always be the potential for edge cases where we fallback on op-geth blocks because of a large influx of orphaned transactions. 

(Noting there are ways to prevent orphaned transactions at the transaction proxy level)

But we feel this is a simple lever we can pull on mainnet to fallback to canonical block building when necessary.

0xOsiris:
Also not sure why CI is failing - it works fine locally 😅

avalonche:
kurtosis seems pretty flaky on CI, also takes quite a while to run

avalonche:
any pointers for fixing kurtosis? it is failing on my pr too

0xOsiris:
> any pointers for fixing kurtosis? it is failing on my pr too

I'll look into it today

Issue #181, opened 2025-04-17
RFC: Rollup Boost HA Design


0xKitsune:
The protocol team at TFH has been exploring high availability designs for Rollup Boost and put together a design doc outlining a few options we’ve considered. Our goal is to align on an approach that’s compatible with the existing OP Stack HA deployment and prioritizes the liveliness and fault tolerance guarantees it offers.

Our mental model is that Rollup Boost should be maximally simple, robust and inherit all of the guarantees from the existing sequencer architecture.

We are currently running Rollup Boost on World Chain Sepolia using a 1:1 Rollup Boost <-> Builder setup. While this is our current approach, this doc isn’t meant to be prescriptive but rather to kick off a discussion on the best way to run Rollup Boost in an HA configuration.

Feel free to let us know if anything needs clarification or if there’s something specific that would be helpful to add. Along with the other L2s planning to run Rollup Boost, we’d love to get OP’s thoughts on a design direction and how to continually iterate to align with the broader vision for the OP Stack.

[Here is the link to the design document](https://github.com/worldcoin/rollup-boost/blob/rollup-boost-ha/docs/rollup-boost-ha.md)

dmarzzz:
Adding HA Technical Design Document from @angel-ding-cb here https://hypnotic-ash-de1.notion.site/TDD-Rollup-Boost-Integration-with-HA-Sequencer-1d0c9d820ca380348f21e44a5442feaf

alfonso-op:
Thanks for drafting, some thoughts:

* The simplicity of the 1:1 rollup-boost <> builder design is appealing. FWIW we will be adopting a similar architecture for Interop (associating a supervisor with each sequencer instance), so the ability to treat each sequencer instance as a single self-contained deployable unit is a nice property and lends to operational simplicity
* One other edge case failure scenario that should be considered for the 1:1 design - if all builders are unhealthy, a sequencer must still be selected as leader but should leverage the default execution client
* The Builder Load Balancer design seems preferable to the n Block Builders design - supports horizontal scaling of the builder without requiring modifying the config of each rollup-boost instance, improved separation of concerns (isolation of routing and block selection logic)
[...]

PR #185, opened 2025-04-21
RFC: Rollup Boost HA Design

0xKitsune:
Opening this PR to migrate the design doc discussion from #181 so reviewers can more easily leave comments and suggest edits inline.


docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+With the introduction of `rollup-boost`, an additional component is introduced that sits in-between `op-node` and `op-geth` that forwards Engine API requests to an external builder.
+
+This design document outlines the architecture, components, and failure strategies required for HA `rollup-boost`. The proposed design prioritizes fault tolerance, liveliness, horizontal scalability, and minimal failover time while maintaining upstream compatibility with `op-conductor`.
+
+## Goals
+- Explore HA designs for `rollup-boost` prioritizing liveliness, fault tolerance and horizontal scalability for external block builders.
+- Maintain compatibility with `op-conductor` and its sequencing assumptions.
+
+## Non Goals
+- Define how Flashblocks are handled, consumed or streamed to the network.
```
0x00101010, line 35: This is actually a very important part that should be included in this design.

Think about one situation, if say in #2 load balancing solution below, what if get_payload call returns from a different builder that streams to public? This will cause flashblock <-> actual block inconsistencies which we aim to avoid as much as possible.

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+
+This approach is operationally simple and relies on the same liveness and fault tolerance guarantees already provided by the OP Stack's sequencer HA setup. Currently, `op-conductor` health checks only evaluate the state of the default execution client, and are unaware of builder health. In the event that the builder is down, `op-conductor` will not switch over to another sequencer instance. Further, if more than one builder is down, `op-conductor` will not currently know if the builder associated with the incoming leader is healthy or unhealthy. One potential solution would be to point `op-conductor` at `rollup-boost` for health checks which could incorporate both the builder and default execution client health. Alternatively `op-conductor` could feature health checks for `rollup-boost` in addition to the existing health checks.
```
0x00101010, line 51: https://github.com/flashbots/rollup-boost/pull/142 could be an interesting solution as described here

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+### Failure Scenarios
+
+<details>
+<summary> Failure Scenarios</summary>
+
+Below is a high level summary of how each failure scenario is handled. All existing failure modes assumed by upstream `op-conductor` are maintained:
+
+| Failure Scenario | Category | Scenario and Solution |
+| --- | --- | --- |
+| Leader Sequencer Fails | Sequencer Failure | Conductor will detect sequencer failure and start to transfer leadership to another node, which will start sequencing instead. Currently,`op-conductor` will not know if the newly elected sequencer has a healthy builder. In the event that `op-conductor` selects a sequencer with an unhealthy builder, manual intervention is needed to either switch the leader or recover the builder. Once the originally failed sequencer is back online, it will join the sequencer set as a follower. |
```
0x00101010, line 104: depending on if we ask conductor to monitor builder health or not, if we did as described above, then op-conductor will know that the newly elected leader is not healthy and then continue to transfer leader until it finds a healthy one (if it was able to).

In the case that we don't ask conductor to monitor builder health, yeah, it'll treat the new sequencer as healthy if its local builder is healthy. Which is the point (conductor cannot act on information it does not know).

So if we want to avoid manual intervention here, we'll need to add builder health check logic to conductor

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+| Leader `rollup-boost` Fails | Rollup Boost Failure | Leader sequencer `rollup-boost` fails, causing `op-conductor`s sequencer health checks to fail, notifying conductor to elect a new leader. This failure mode is the same as a typical leader sequencer failure. Once the sequencer `rollup-boost` recovers, the sequencer will rejoin the sequencer set as a follower. |
```
0x00101010, line 105: to be clear, the sequencer is always in the cluster as follower, it's just that it won't start sequencing since it's not considered healthy

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+<summary>Pros/Cons</summary>
+
+#### Pros
+
+- Simple, easy to operate and monitor
+- No coordination required between `n` builders
+
+#### Cons
+
+- No builder failover/redundancy
```
0x00101010, line 125: I understand the logic behind this statement, but I believe this actually provides builder redundancy, and failover (if we monitor builder health)

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+- Resource intensive for chains that do not want to run 3 builders
+
+</details>
+
+<br/>
+
+## `n` Block Builders
+
+In this design, each `rollup-boost` instance can be configured with `n` builders. When an FCU with payload attributes is received, it is broadcast to all builders, each of which begins building a block. Once `rollup-boost` receives a `get_payload` request from `op-node`, it multiplexes the request to all builders. Each builder returns its built payload, and if a payload isn't returned within `n` ms, it is ignored. A block selection policy is applied across the successful responses to choose the final payload (eg. selecting the block with the highest gas used). The selected payload is then validated with via a `new_payload` request sent to the sequencer's default execution client. In the event that the selected block is invalid, the next best block is selected and validated with the default execution client instead. If all builder blocks are invalid, then the default execution client's block is selected.
```
0x00101010, line 134: Unfortunately this solution does not work mainly due to that we have to decide which builder's payload you're gonna potentially accept, otherwise, it's not feasible to decide who's flashblocks to stream to the public

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+#### Cons
+
+- No strict guarantee builders only receive one FCU with payload attributes.
+- Block selection logic adds complexity to `rollup-boost`  
+
+</details>
+
+<br/>
+
+## Builder Load Balancer
```
0x00101010, line 228: same issue as the `n` builder solution, we need to choose builder deterministically (at least within a block) so that we know the exact flashblocks to stream

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+
+Importantly, this design requires no changes to `op-conductor`, allowing Conductor simply query `op-node` for health checks, keeping builder health and state completely out of the leader election process.
+
+One potential issue with this design is that `op-conductor` does not strictly guarantee there will only be one active sequencer at a time. Specifically, there are edge cases where the `stopSequencer` call is not received in time (or if the call fails) by the previous leader’s `op-node` after a conductor failover. This edge case is acknowledged in the [op-conductor codebase and currently considered acceptable](https://github.com/ethereum-optimism/optimism/blob/273fd6abd4a92824e05f3daa9322b0c8add9e843/op-conductor/conductor/service.go#L766). The [Engine API](https://github.com/ethereum/execution-apis/blob/main/src/engine/common.md#load-balancing-and-advanced-configurations) spec states it is undefined behavior when multiple consensus clients concurrently control the head of a single execution client. Some clients may tolerate this, but it’s not standardized and could lead to unpredictable behavior.
```
0x00101010, line 188: This is a "normal" case that HA system needs to handle. Think of the situation where conductor is down, leadership transfers and now we have 2 active sequencers. The system is able to handle this case to tolerate 2 active sequencers without affecting entire network's health

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,328 @@
...
+- Define how pending transactions are relayed/peered across the builders/sequencer execution clients.
+- Monitoring / alerting strategies. This can be specified in a separate document once an architecture is solidified.
+
+# Designs
+
+The following designs build on the existing HA sequencer setup by introducing a `rollup-boost` instance between each `op-node` and its local `op-geth` instance.
+
+## 1:1 Rollup Boost to Builder Deployments
+
+In this design, each `rollup-boost` instance is configured with a single external builder and default execution client. When `op-node` sends an FCU containing payload attributes, `rollup-boost` forwards the request to both the default execution client and its paired builder. Upon receiving a `get_payload` request from `op-node`, `rollup-boost` queries both the execution client and the builder. If the builder returns a payload, it is validated via a `new_payload` request sent to the default execution client. If the builder payload is invalid or unavailable, `rollup-boost` falls back to the execution client's payload.
```
0x00101010, line 45: In addition to the current setup, another feature will need to be added here.

We'll need to allow conductor to proxy flashblocks web socket connections / results to the public (based on if the current node is leader or not). This is a common pattern used in conductor, proxy any interaction with the system that strictly requires to interact with the leader.

0x00101010: One thing to note here is that during leadership transfer, there could be flashblocks reorg in the current design.

And if we want to prevent reorgs, we'll need to consider committing flashblocks to the raft consensus for consistency during leadership transfer, and that is second phase of our planned work

docs/rollup-boost-ha.md
0xKitsune: Yes agreed! With #142 merged, I updated the design doc to reflect the new health checks exposed by `rollup-boost` which now reports healthy, partially healthy, or unhealthy states.

docs/rollup-boost-ha.md
0xKitsune: > So if we want to avoid manual intervention here, we'll need to add builder health check logic to conductor
Yes I think we should avoid manual intervention here. The current plan is to [add `rollup-boost` monitoring to `op-conductor`](https://github.com/ethereum-optimism/optimism/pull/15316). With #142 merged, this is now unblocked.

docs/rollup-boost-ha.md
0xKitsune: Agreed, the design doc has been updated to remove this.

docs/rollup-boost-ha.md
0xKitsune: Yes we are aligned on this. After an initial round of feedback, the design doc has been updated to focus on 1:1 Rollup Boost <-> Builder deployments.

docs/rollup-boost-ha.md
0xKitsune: The design doc has been updated to focus on 1:1 Rollup Boost <-> Builder deployments which ensures that there is only one `op-node` sending FCUs with payload attributes to the EL.

docs/rollup-boost-ha.md
0xKitsune: Thanks for flagging, this has been updated.

docs/rollup-boost-ha.md
0xKitsune: Sounds good, as mentioned in a previous reply the design doc has been updated to focus on 1:1 Rollup Boost <-> Builder HA deployments.

jelias2:
1. Flashblocks TX Ingress. 
  - Please correct this statement if I'm wrong, but flashblocks receipts are provided to the end user when the tx is sent directly to the builder. If this is the case then we may want to discuss how to design ingress solutions which target the active builder yet do not overload it.

2. Flashblock Data Exports
 - Base has built the flashblocks-websocket-proxy. Ideally in a production setup this data should be exportable by every sequencer in a conductor set. Solutions may need to be built for this as well

docs/rollup-boost-ha.md
0x00101010: should we update this now? I think we need to take this into consideration when designing HA solutions for flashblocks

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+|:----------|:--------------|
+| Builder is synced and both execution client and builder return payloads | `200 OK` (Healthy) |
+| Builder is out of sync| `206 Partial Content` (Partially Healthy) |
+| Builder fails to return payload on `get_payload` request | `206 Partial Content` (Partially Healthy) |
+| Execution client fails to return payload on `get_payload` request | `503 Service Unavailable` (Unhealthy) |
+
+`op-conductor` should query the `/healthz` endpoint exposed by Rollup Boost in addition to the existing execution client health checks. Health should be interpreted as follows:
+
+- `200 OK` (Healthy): The node is fully healthy and eligible for leadership.
+- `206 Partial Content` (Partially Healthy): The node is degraded but may be considered for leadership if no fully healthy candidates are available.
```
0x00101010, line 123:
```suggestion
- `206 Partial Content` (Partially Healthy): The node is degraded but may be considered for leadership if configured by operator
```

docs/rollup-boost-ha.md
0xKitsune: For clarity, on the design review call last week we aligned on focusing this doc on an HA design that is forward compatible with Flashblocks without defining the full Flashblocks HA specifics here. 

The plan is to fast follow with a separate document that extends the HA design detailing Flashblocks HA behavior including `op-conductor` integration and consistency guarantees. Happy to update and clarify to make this more explicit. Let me know if you have any additional thoughts on this.

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+- `206 Partial Content` (Partially Healthy): The node is degraded but may be considered for leadership if configured by operator
+- `503 Service Unavailable` (Unhealthy): The node is unhealthy and must be excluded from leadership.
+
+During normal operation and leadership transfers, `op-conductor` should prioritize candidates in the following order:
+
+1. Prefer nodes reporting `200 OK`.
+2. If no fully healthy nodes are available, select from nodes reporting `206 Partial Content`.
+3. Nodes reporting `503 Service Unavailable` must not be selected as leader.
+
+Rollup Boost instances that are not actively sequencing rely exclusively on the builder sync check to report health, as they are not producing blocks. This behavior mirrors the existing `op-conductor` health checks for inactive sequencers and ensures readiness during failover without compromising network liveness guarantees.
```
0x00101010, line 132: Can you please expand on that? Why don't we want to have follower node report local sequencer health?

For example, if the follower got elected leader, builder is in sync, but the local sequencer is out of sync, we'd want to transfer leader to another healthy one, right?

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+
+<br>
+
+## Failure Scenarios
+
+Below is a high level summary of how each failure scenario is handled. All existing failure modes assumed by upstream `op-conductor` are maintained:
+
+| Failure Scenario | Category | Scenario and Solution |
+| --- | --- | --- |
+| Leader Sequencer Fails | Sequencer Failure | Conductor will detect sequencer failure and start to transfer leadership to another node, which will start sequencing instead. `op-conductor` will prioritize electing a leader with full health however if there are no sequencers that are fully healthy, a sequencer with partial health will be selected. Once the originally failed sequencer is back online, it will join the sequencer set as a follower. |
```
0x00101010, line 239: same as above, suggest not to ask conductor to prioritize certain sequencers for the reason mentioned

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+| Leader `rollup-boost` Unhealthy | Rollup Boost Failure | Leader sequencer `rollup-boost` becomes unhealthy, causing `op-conductor`s sequencer health checks to fail. If `op-conductor` is configured to failover in the event that `rollup-boost` is partially healthy conductor will elect a new leader. This failure mode is the same as a typical leader sequencer failure. Once the sequencer recovers, it will continue to participate in the cluster as a follower. If `op-conductor` is configured not to failover in the event of `rollup-boost` partial health, the sequencer will continue to publish the default exectuion client's blocks until the builder has recovered back to full health. |
```
0x00101010, line 240: maybe rename rollup-boost failure to rollup-boost partial healthy? 

Cuz if rollup-boost is borked / completely failed, we should transfer leadership

docs/rollup-boost-ha.md
0xKitsune: > Can you please expand on that? Why don't we want to have follower node report local sequencer health?

For clarity, this section isn’t proposing that `op-conductor` should rely solely on the builder health checks. The existing sequencer health checks would still be used to evaluate sequencer status as they are today. This is additive, what’s described here introduces additional health checks focused specifically on assessing builder health alongside the existing ones. Follower nodes would still report local sequencer health status.

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+| Follower `rollup-boost` Unhealthy | Rollup Boost Failure | Follower sequencer `rollup-boost` becomes unhealthy. The leader sequencer is unaffected. Once the sequencer recovers, it will continue to participate in the cluster as a follower.|
```
0x00101010, line 241: same here

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+
+- **Payload Production**:  
+  During each `get_payload` request, Rollup Boost will verify payload availability from both the builder and the execution client. If the builder fails to deliver a payload, Rollup Boost will report partial health. If the execution client fails to deliver a payload, Rollup Boost will report unhealthy.
+
+`op-conductor` should also be configurable in how it interprets health status for failover decisions. This allows chain operators to define thresholds based on their risk tolerance and operational goals. For example, operators may choose to maintain leadership with a sequencer reporting `206 Partial Content` to avoid unnecessary fail overs or they may configure `op-conductor` to immediately fail over when any degradation is detected. This flexibility allows the chain operator to configure a failover policy that aligns with network performance expectations and builder reliability.
```
0x00101010, line 109: Thinking about this a little bit more, curious is there any specific reason (other than convenience that rollup-boost already connects to both builder and default EL) that we do the health check in rollup-boost, not conductor?

The potential drawback here is that due to potentially different configurations or actual health check cadence, there might be a delay in when builder is reported unhealthy and conductor knows it and starts the leadership transfer.

docs/rollup-boost-ha.md
0x00101010: I think we're aligned, but should we update the message to remove `exclusively`? Maybe change to `additionally`?

0xKitsune: Sounds good, I have updated this section to reflect this. Let me know if you think it should be adjusted further.

> Rollup Boost instances that are not actively sequencing rely on the builder sync check to report health, as they are not producing blocks.  This behavior mirrors the existing `op-conductor` health checks for inactive sequencers and ensures readiness during failover without compromising network liveness guarantees. Note that `op-conductor` will still evaluate existing sequencer health checks to determine overall sequencer health.

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+
+- `200 OK` (Healthy): The node is fully healthy and eligible for leadership.
+- `206 Partial Content` (Partially Healthy): The node is degraded but may be considered for leadership if no fully healthy candidates are available.
+- `503 Service Unavailable` (Unhealthy): The node is unhealthy and must be excluded from leadership.
+
+During normal operation and leadership transfers, `op-conductor` should prioritize candidates in the following order:
+
+1. Prefer nodes reporting `200 OK`.
+2. If no fully healthy nodes are available, select from nodes reporting `206 Partial Content`.
+3. Nodes reporting `503 Service Unavailable` must not be selected as leader.
```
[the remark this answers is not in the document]
0xKitsune: Makes sense, I have updated this section to the following:

> 1. Prefer nodes reporting `200 OK`.
> 2. Nodes that return `503 Service Unavailable` are treated as unhealthy and must not be eligible for sequencer leadership. `op-conductor` should offer a configuration option to treat nodes returning `206 Partial Content` as either healthy or unhealthy.

Let me know if this aligns with your thinking. If not feel free to suggest how to update.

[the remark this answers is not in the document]
0x00101010: Looks good, thank you!

docs/rollup-boost-ha.md
0xKitsune: I've updated the failure scenarios to clarify behavior for both partial and full `rollup-boost` failure, as well as local execution client and builder failures.

docs/rollup-boost-ha.md
0xKitsune: I responded to the prev comment but copying here for visibility. 

Failure scenarios have been updated clarify behavior for both partial and full `rollup-boost` failure as well as local execution client and builder failures.

docs/rollup-boost-ha.md
0xKitsune: > Thinking about this a little bit more, curious is there any specific reason (other than convenience that rollup-boost already connects to both builder and default EL) that we do the health check in rollup-boost, not conductor?

Our current thinking is that `rollup-boost` health checks should evaluate both if the builder is synced as well as if it is producing valid payloads.

By placing the builder health check in `rollup-boost`, we can assess payload health in real time during `get_payload` calls, ensuring the builder is not only responsive but also producing valid blocks. Each builder payload is validated via a `newPayload` call to the local execution client, confirming both construction and validity before marking the builder as healthy.

While `op-conductor` could perform basic sync checks, this setup allows the health check to ensure block production/validity, giving us a stronger signal of health than sync checks alone. Let me know if you have any thoughts on this.

0x00101010: Cool, I think this makes sense, one potential suggestion here was to mention (maybe not exactly here) that to minimize health check delay, would suggest to clearly consider / configure matching health check intervals for both conductor and rollup-boost

0x00101010 (approved):
LGTM!

jelias2 (approved):
lgtm

0xForerunner (approved):

karankurbur (approved):

docs/rollup-boost-ha.md
SozinM: There is little contradiction in the failover section because of flashblocks.
In https://flashbots.notion.site/Draft_Flashblocks_Eth_OPstack-md-1d66b4a0d876808493c9dcc4428ccad7#1d66b4a0d87680c8b550d356ad113318 it says that we prioritize preconfirmations, so if leader would become partially healthy (builder is down) we will still use block built by builder. In https://github.com/flashbots/rollup-boost/pull/185/files#diff-df3270fd8794e55e7c8f29d84e2605dc61fecaf32696ac4788661c0038e19a6fR242 it says that we will choose block from the EL.

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+
+<br>
+
+## Failure Scenarios
+
+Below is a high level summary of how each failure scenario is handled. All existing failure modes assumed by upstream `op-conductor` are maintained:
+
+| Failure Scenario | Category | Scenario and Solution |
+| --- | --- | --- |
+| Leader Sequencer Execution Client Fails | Sequencer Failure | `op-conductor` will detect an unhealthy status from both `rollup-boost` and pre-existing sequencer health checks, causing Conductor to elect a new leader. Once the default exectuion client has recovered, `rollup-boost` will update it's health status to `200` and the sequencer will continue operating normally as a follower. |
```
odyslam, line 238:
```suggestion
| Leader Sequencer Execution Client Fails | Sequencer Failure | `op-conductor` will detect an unhealthy status from both `rollup-boost` and pre-existing sequencer health checks, causing Conductor to elect a new leader. Once the default execution client has recovered, `rollup-boost` will update it's health status to `200` and the sequencer will continue operating normally as a follower. |
```

docs/rollup-boost-ha.md
```diff
@@ -0,0 +1,245 @@
...
+| Leader Builder Falls Out of Sync | Builder Failure | The builder associated with the sequencer leader falls out of sync with the chain head. `rollup-boost` will detect the unsynced state via the background health checks and downgrade its health status to partial. This will result in `rollup-boost` ignoring builder payloads and selecting the default execution client payload for block until the builder is resynced. If `op-conductor` is configured to failover upon partial `rollup-boost` health, a new leader will attempt to be elected. Once the builder recovers, `rollup-boost` will update its health to `200` and resume with normal operation. |
```
odyslam, line 243: Given that external block building is the intended effect for using Rollup-Boost, I would assume that failover on partial failure should be the default behavior. 

Maybe it should work something like this:
1. Block builder fails (cluster A)
2. Default is switched to Sequencer EL and produces the next block ( cluster A )
3. In the meantime, failover has started to switch to ( cluster B ), so there is an upper bound of x blocks (ideally 1) that are produced by a sequencer, before switching to another block builder.

PR #186, opened 2025-04-24
Remove extra log from rollup_boost.log

ferranbt:
This PR removes an extra log line in the integration tests which was appending to `rollup_boost.log` the logs from all the extra services.


0xForerunner (approved):
Ah nice! Good catch, TY.

PR #192, opened 2025-04-29
Process engine requests sequentially

ferranbt:
This PR changes how we handle engine API requests by processing them sequentially rather than concurrently. The changes implement a message-passing architecture with an EngineHandler that forwards requests to the main RollupBoostServer through a channel.

Motivation:

- Alignment with EL expectations: Execution Layer expects to receive and process messages in sequence, not in parallel
- Concurrency control: Prevents race conditions when the Consensus Layer sends many requests in a short timeframe

TODO: Write the success condition in the e2e test.


ferranbt:
Closing in favour of #201

PR #197, opened 2025-04-30
Add e2e tests for Isthmus fork

### LINKED ISSUE #194: Update tests with v4 payload calls
Isthmus is fast approaching, all the tests are still on v3, something we would like to update

ferranbt:
Closes #194 

- Introduced a new Genesis struct and refactored genesis file handling to support configurable Isthmus hardfork parameters including isthmusTime and pragueTime
- Enhanced RollupBoostTestHarnessBuilder and SimpleBlockGenerator with methods for Isthmus-specific configurations and logic for transitioning between protocol versions (V3 to V4)
- Added integration tests specifically for validating Isthmus hardfork scenarios, both from genesis and during transition periods


tests/simple_isthmus_transition.rs
```diff
@@ -0,0 +1,15 @@
+use common::RollupBoostTestHarnessBuilder;
+
+mod common;
```
ferranbt, line 3: I had to create separated files for each one because there are some global setters in Rollup-boost that would crash if they were set twice (random, tracing, metrics). I will leave for a follow up PR to simplify this if we want to have all these simple tests in a single file.

tests/simple.rs
```diff
@@ -4,16 +4,12 @@ mod common;
 
 #[tokio::test]
 async fn test_integration_simple() -> eyre::Result<()> {
```
avalonche, line 6: what's the difference between this test and `test_integration_simple_isthmus_transition`? feel the tests are the same

avalonche (approved):

tests/simple.rs
ferranbt: This one starts with a chain that is v3 and test_integration_simple_isthmus_transition transitions from a chain that uses v3 to one that uses v4 at block 5.

PR #198, opened 2025-04-30
feat: simplify ExecutionMode

### LINKED ISSUE #196: Simplify `ExecutionMode`
Rollup Boost supports four `ExecutionMode` variants that seek to control how `rollup-boost` interacts with the external builder during block production. Currently the four modes are `Enabled`, `DryRun`, `Fallback`, and `Disabled` where the differences between `DryRun` and `Fallback` are minimal. Execution modes could be simplified to the following:

- `Enabled`
  - `rollup-boost` forwards all Engine API requests to both the builder and default execution client.
  - Optimistically selects the builder’s payload for validation and block publication.
  - Falls back to the local execution client *only* if the builder fails to produce a payload or the payload is invalid.
  - Default setting for normal external block production.

- `DryRun`  
  - `rollup-boost` forwards all Engine API requests to both the builder and default execution client.
  - Builder payloads are validated with the local execution client but the default execution client block will always be returned to `op-node` to propagate to the network.
  - Useful during deployments, dry runs, or to validate builder behavior without publishing builder blocks to the network.

- `Disabled`
  - `rollup-boost` does not forward any Engine API requests to the builder.
  - Block construction is handled exclusively by the default execution client.
  - Useful as an emergency shutoff switch in the case of critical failures/emergencies.

```rust
pub enum ExecutionMode {
    /// Forward Engine API requests to the builder, validate builder payloads and propagate to the network 
    Enabled,
    /// Forward Engine API requests to the builder, validate builder payloads but
    /// fallback to default execution payload
    DryRun,
    // Do not forward Engine API requests to the builder 
    Disabled,
}
```

varun-doshi:
Fixes #196 

Changes made:
- Removed `Fallback` mode from `ExecutionMode`
- `DryRun` mode will get both `l2_payload` and `builder_payload`, but will always send the `l2_payload`. This was different previously where in DryRun, the call to `builder_payload` would return an error.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -628,7 +628,10 @@ impl RollupBoostServer {
                     Ok(payload)
                 }
 
-                Err(e) => Err(e.into()),
+                Err(e) => {
+                    self.probes.set_health(Health::ServiceUnavailable);
+                    Err(e.into())
+                }
             };
         }
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -611,17 +611,16 @@ impl RollupBoostServer {
             return match l2_fut.await {
                 Ok(payload) => {
                     self.probes.set_health(Health::Healthy);
-                    tracing::Span::current()
-                        .record("payload_source", PayloadSource::L2.to_string());
-                    counter!("rpc.blocks_created", "source" => PayloadSource::L2.to_string())
-                        .increment(1);
+                    let context = PayloadSource::L2;
+                    tracing::Span::current().record("payload_source", context.to_string());
+                    counter!("rpc.blocks_created", "source" => context.to_string()).increment(1);
 
                     let execution_payload = ExecutionPayload::from(payload.clone());
                     info!(
                         message = "returning block",
                         "hash" = %execution_payload.block_hash(),
                         "number" = %execution_payload.block_number(),
-                        context = PayloadSource::L2.to_string(),
+                        %context,
                         %payload_id,
                     );
 
--- a/src/server.rs
+++ b/src/server.rs
@@ -603,36 +603,11 @@ impl RollupBoostServer {
         payload_id: PayloadId,
         version: Version,
     ) -> RpcResult<OpExecutionPayloadEnvelope> {
-        let l2_client_future = self.l2_client.get_payload(payload_id, version);
-        let builder_client_future = Box::pin(async move {
-            if let Some(cause) = self.payload_trace_context.trace_id(&payload_id) {
-                tracing::Span::current().follows_from(cause);
-            }
-
-            if !self.payload_trace_context.has_builder_payload(&payload_id) {
-                // block builder won't build a block without attributes
-                info!(message = "builder has no payload, skipping get_payload call to builder");
-                return RpcResult::Ok(None);
-            }
-
-            let builder = self.builder_client.clone();
-            let payload = builder.get_payload(payload_id, version).await?;
-
-            // Send the payload to the local execution engine with engine_newPayload to validate the block from the builder.
-            // Otherwise, we do not want to risk the network to a halt since op-node will not be able to propose the block.
-            // If validation fails, return the local block since that one has already been validated.
-            let _ = self
-                .l2_client
-                .new_payload(NewPayload::from(payload.clone()))
-                .await?;
-
-            Ok(Some(payload))
-        });
-
         let execution_mode = self.execution_mode();
+        let l2_client_future = self.l2_client.get_payload(payload_id, version);
 
-        // if mode is `dry_run` dont want to wait for the builder payload
-        let (payload, context) = if execution_mode.is_dry_run() {
+        // if mode is `disabled` dont await the builder payload
+        let (payload, context) = if execution_mode.is_disabled() {
             let result = match l2_client_future.await {
                 Ok(payload) => {
                     self.probes.set_health(Health::Healthy);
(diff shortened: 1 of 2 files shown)


src/server.rs
```diff
@@ -653,7 +634,7 @@ impl RollupBoostServer {
             (Ok(Some(builder)), Ok(l2_payload)) => {
                 // builder successfully returned a payload
                 self.probes.set_health(Health::Healthy);
-                if self.execution_mode().is_fallback_enabled() {
+                if self.execution_mode().is_dry_run() {
```
ferranbt, line 637: It would be nice if we could make it not wait for the other future to finish so that it does not affect the normal flow execution.

varun-doshi: So you mean if its a `dry_run` it should not have to wait for the `builder_payload` ?
The issue stated that even in `dry_run` it should get and validate both `builder` and `l2` payloads.

Correct me if I'm wrong here

ferranbt: Saw your latest PR but I might be good either way, @0xKitsune WDYT?

0xKitsune: The initial thinking for `ExecutionMode::DryRun` was to await the builder payload so that it could be validated via a `new_payload()` call to the default execution client. I think this approach is preferred so that operators can ensure the builder is healthy, synced and producing valid blocks before switching to `ExecutionMode::Enabled`.

0xKitsune:
Thanks for opening this PR, in addition to the comments above can we also ensure that Rollup Boost does not forward Engine API requests to the builder when ExecutionMode is set to `Disabled`?

src/server.rs
varun-doshi: So then we want to get the builder payload even in `dry_run` right? Fetch and validate it but dont send it. Always send l2 payload?

varun-doshi:
Latest Commit makes the following changes:

- In Disabled Mode, dont await the builder payload. Only get the `l2_payload` and send that
- In Enabled/ Dry Run mode, get both builder + l2 payload.
- In `dry_run` mode:
  - if builder + l2 is okay, send the `l2 payload` with `Healthy` signal
  - if builder is none/error and l2 is okay, send `l2 payload` with `PartialContent` signal
- In `enabled` mode:
  - if builder + l2 is okay, send the `builder` payload with `Healthy` signal
  - if builder is none/err and l2 is okay send `l2` with `PartialContent` signal
  - If both are none/err, send `ServiceUnavailable` signal

src/server.rs
```diff
@@ -610,66 +603,79 @@ impl RollupBoostServer {
...
+                    Ok((l2, PayloadSource::L2))
+                }
+                (_, Err(e), _) => {
+                    // l2 failed to return a payload
+                    self.probes.set_health(Health::ServiceUnavailable);
+                    Err(e)
+                }
+            };
+            result
         }?;
```
0xKitsune, line 679: I think we can simplify the structure a bit further and avoid the need for verbose pattern matching. Something like the following should work well:

```rust
    async fn get_payload(
        &self,
        payload_id: PayloadId,
        version: Version,
    ) -> RpcResult<OpExecutionPayloadEnvelope> {
        let l2_payload_fut = self.l2_client.get_payload(payload_id, version);

        // If exectuion mode is disabled, return the l2 payload without sending
        // the request to the builder
        if self.execution_mode().is_disabled() {
            return match l2_payload_fut.await {
                Ok(payload) => {
                    self.probes.set_health(Health::Healthy);
                    tracing::Span::current().record("payload_source", "L2");
                    counter!("rpc.blocks_created", "source" => "L2").increment(1);

                    let execution_payload = ExecutionPayload::from(payload.clone());
                    info!(
                        message = "returning block",
                        "hash" = %execution_payload.block_hash(),
                        "number" = %execution_payload.block_number(),
                        context = "L2",
                        %payload_id,
                    );

                    Ok(payload)
                }

                Err(e) => Err(e.into()),
            };
        }

        // Forward the get payload request to the builder
        let builder_fut = async {
            if let Some(cause) = self.payload_trace_context.trace_id(&payload_id) {
                tracing::Span::current().follows_from(cause);
            }
            if !self.payload_trace_context.has_builder_payload(&payload_id) {
                info!(message = "builder has no payload, skipping get_payload call to builder");
                return RpcResult::Ok(None);
            }

            // Get payload and validate with the local l2 client
            let payload = self.builder_client.get_payload(payload_id, version).await?;
            let _ = self
                .l2_client
                .new_payload(NewPayload::from(payload.clone()))
                .await?;

            Ok(Some(payload))
        };

        let (l2_payload, builder_payload) = tokio::join!(l2_payload_fut, builder_fut);

        // Evaluate the builder and l2 response and select the final payload
        let (payload, context) = {
            let l2_payload =
                l2_payload.inspect_err(|_| self.probes.set_health(Health::ServiceUnavailable))?;
            self.probes.set_health(Health::Healthy);

            if let Ok(Some(payload)) = builder_payload {
                if self.execution_mode().is_dry_run() {
                    (l2_payload, PayloadSource::L2)
                } else {
                    (payload, PayloadSource::Builder)
                }
            } else {
                self.probes.set_health(Health::PartialContent);
                (l2_payload, PayloadSource::L2)
            }
        };

       // --snip--
    }
```

varun-doshi: Ahh yes, you're right. This looks much better

varun-doshi:
There's a failing test which says `Unknown context: "l2"` which is weird because there's a match for `l2`

0xKitsune:
>There's a failing test which says Unknown context: "l2" which is weird because there's a match for l2 in the tests/common/mod.rs file

The test is failing because the way `context` is injected into the `info!` call when execution mode is disabled differs from when execution mode is enabled. This mismatch causes the test to fail with `Unknown context: "l2"`. Aligning the logging format between the two paths will resolve the issue.

0xKitsune (approved):

0xKitsune:
Thanks for the PR, LGTM! 

@ferranbt @avalonche any additional thoughts? I noticed that the Kurtosis CI keeps failing due to an issue with how Contender is being run. We can open a separate PR to fix this.

danyalprout:
@0xKitsune opened a PR to fix the stress test CI https://github.com/flashbots/rollup-boost/pull/204

PR #199, opened 2025-04-30
Add log for fcu

ferranbt:
This PR adds a log to print the specific type of FCU call and whether the txpool is enabled for the one that triggers block building.

To give some extra context, during the block building lifecycle there are two FCU calls, one for block building and one after the block has been included on chain. This logs helps to differentiate both cases.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -314,7 +314,15 @@ impl EngineApiServer for RollupBoostServer {
         fork_choice_state: ForkchoiceState,
         payload_attributes: Option<OpPayloadAttributes>,
     ) -> RpcResult<ForkchoiceUpdated> {
-        info!("received fork_choice_updated_v3");
+        if let Some(attr) = payload_attributes.as_ref() {
+            info!(
+                message = "received fork_choice_updated_v3 with payload attributes",
+                "use_tx_pool" = !attr.no_tx_pool.unwrap_or_default()
+            );
+        } else {
+            info!("received fork_choice_updated_v3");
+        }
+
         // First get the local payload ID from L2 client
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -314,7 +314,15 @@ impl EngineApiServer for RollupBoostServer {
...
-        info!("received fork_choice_updated_v3");
+        if let Some(attr) = payload_attributes.as_ref() {
+            info!(
+                message = "received fork_choice_updated_v3 with payload attributes",
+                "use_tx_pool" = !attr.no_tx_pool.unwrap_or_default()
+            );
+        } else {
+            info!("received fork_choice_updated_v3");
+        }
+
```
avalonche, line 325:
```suggestion
        let has_attribute = payload_attributes.is_some();
        let no_tx_pool = payload_attributes
            .as_ref()
            .is_some_and(|attr| attr.no_tx_pool.unwrap_or_default());
        info!(
                message = "received fork_choice_updated_v3",
                "has_attributes" = has_attributes,
                "use_tx_pool" = !no_tx_pool,
            );
```

ferranbt: I want to have two messages because `use_tx_pool` is only valid if `has_attributes` is set so we are over logging when it is an fcu without attributes. I feel it is more clear on the logs what is happening in the internal lifecycle if we differentiate the messages.

avalonche (approved):

src/server.rs
avalonche: having differentiated logging messages is fine, I feel that adding tags it is easier to filter the logs

Issue #202, opened 2025-05-02
Migrate `websocket-proxy` to this repo


ferranbt:
Migrate the [flashblocks-websocket-proxy](https://github.com/base/flashblocks-websocket-proxy) repo here.

(cc @danyalprout)

danyalprout:
@ferranbt the PR for this should be good to go, I've preserved the history from the original repo -- so only a small handful of commits exist at the end that haven't been review

PR #203, opened 2025-05-06
chore: move websocket proxy

danyalprout:
### Description
Initial pass for https://github.com/flashbots/rollup-boost/issues/202

* This PR moves the websocket proxy code from [base/flashblocks-websocket-proxy](https://github.com/base/flashblocks-websocket-proxy) into this repo.
* Added https://github.com/flashbots/rollup-boost/pull/203/commits/2615ec490d6667a32c84ace64275d084d2977b28 to combine the CI Github Actions into the workflow (I'll add the release one separately)
* Added https://github.com/flashbots/rollup-boost/pull/203/commits/7e9bdfcc0e2e01f6bbf3370ea299a078e86d612b to fix a clippy issue


websocket-proxy/README.md
```diff
@@ -0,0 +1,64 @@
...
+- `--redis-key-prefix` - Prefix for Redis keys (default: `flashblocks`)
+
+Example:
+
+```bash
+docker run ghcr.io/base/flashblocks-websocket-proxy:master \
+  --upstream-ws wss://your-sequencer-endpoint \
+  --redis-url redis://redis:6379 \
+  --global-connections-limit 1000 \
+  --per-ip-connections-limit 10
```
danyalprout, line 54: I'll update the image URLs when I add the docker deployment and have new images

danyalprout:
The CI issue for contender is fixed in https://github.com/flashbots/rollup-boost/pull/204

0xKitsune (commented):
Since we are adding another service independent from `rollup-boost`, I think its best to restructure the repo to use Cargo workspaces. Let me know if you need any help with this.

0xKitsune:
>  Happy to add it as a workspace setup if ya'll prefer

Sounds good, yeah I think a workspace would be best here rather than standalone crates. Workspaces will simplify dependency management across services without duplicating Cargo.toml or needing to keep versions in sync. It also makes the repo a bit more maintainable as more components get added. Lmk if I am overlooking something here but my preference would be to go with workspaces.

.github/workflows/websocket-proxy-ci.yaml
```diff
@@ -0,0 +1,67 @@
+name: Websocket Proxy CI
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+
```
0xKitsune, line 7: Should we constrain this workflow to only run on PRs to `main`?

websocket-proxy/src/integration.rs
```diff
@@ -0,0 +1,308 @@
+mod test {
```
0xKitsune, line 1: Wdyt about moving integration tests to a `tests/` dir? This way the `integration` feature can be removed.

ferranbt:
> cc @ferranbt if I understood, you had a preference to keep them separate/just add a subfolder? (Though I may have misunderstood)

Yes, this was my initial suggestion to keep the PR as lean as possible. If we have to create a workspace we have to restructure everything right? @0xKitsune

0xKitsune:
> Yes, this was my initial suggestion to keep the PR as lean as possible.
Setting up a workspace does not require major restructuring, however happy defer this to a follow up PR to keep changes minimal.

The only change needed is to move the existing crate into a new directory (e.g. rollup-boost/), update the existing `Cargo.toml` to be a workspace member and add a top level `Cargo.toml` with shared dependencies. Either approach works for me, fine to merge first if preferred.

websocket-proxy/src/integration.rs
0xKitsune: Resolving this for now, this can be revisited in a follow up PR.

danyalprout:
I'll me make the workspace changes on this PR. If it results in too large a change we can revert the commit + merge separately.

paranoiacblack:
Hi @danyalprout, is [base/flashblocks-websocket-proxy](https://github.com/base/flashblocks-websocket-proxy) basically closed to changes while this move is happening? If I'd like to make changes to the websocket proxy, should I just fork the branch for this PR instead?

PR #206, opened 2025-05-12
Send FCUs to L2/Builder concurrently

ferranbt:
This PR refactors how Fork Choice Updates (FCUs) are handled in the rollup-boost system to improve clarity and reliability:

Key Changes
- Parallel FCU Processing: Both L2 and Builder FCUs are now sent concurrently. Before, the Builder call would be made sequentially after the L2 call.
- Explicit FCU Handling: Separates logic for FCUs with and without payload attributes for clearer code organization
- **Wait for Both Responses**: For FCUs with payload attributes, we now wait for both L2 and Builder responses before proceeding, fixing a race condition where GetPayload calls could reach the system before FCU calls were completed
- Removed boost-sync Flag: Eliminated the legacy boost-sync option as it's now mandatory for the Builder to function properly

Added test coverage for the case where L2 client fails an FCU call, ensuring we properly return that error and don't progress the chain


src/server.rs
```diff
@@ -314,71 +294,80 @@ impl EngineApiServer for RollupBoostServer {
...
+            .fork_choice_updated_v3(fork_choice_state, payload_attributes.clone());
 
-        let span = tracing::Span::current();
-        if let Some(payload_id) = l2_response.payload_id {
-            span.record("payload_id", payload_id.to_string());
-        }
+        let builder_client = self.builder_client.clone();
+        let builder_client_future = Box::pin(async move {
+            if !should_send_to_builder {
+                return Ok::<Option<ForkchoiceUpdated>, jsonrpsee::types::error::ErrorObject>(None);
```
ferranbt, line 336: Not sure how to get rid of that

tests/execution_mode.rs
```diff
@@ -21,6 +21,7 @@ impl ProxyHandler for CounterHandler {
         _params: Value,
         _result: Value,
     ) -> Pin<Box<dyn Future<Output = Option<Value>> + Send>> {
+        println!("CounterHandler::handle");
```
avalonche, line 24: debug print statements?

src/cli.rs
```diff
@@ -159,17 +159,11 @@ impl Args {
             PayloadSource::Builder,
         )?;
 
-        let boost_sync_enabled = !self.no_boost_sync;
```
avalonche, line 162: should we remove this from .env / cli and docs?

0xKitsune (approved):

PR #211, opened 2025-05-14
fix(auth): use only current timestamp for auth

Dzejkop:
As per the [spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/authentication.md#jwt-claims) this will give a token a vailidity of 60 seconds. Prior to this change the validity would have been 120 seconds.

### DIFF
--- a/src/client/auth.rs
+++ b/src/client/auth.rs
@@ -4,7 +4,7 @@ use http::{HeaderValue, header::AUTHORIZATION};
 use std::{
     iter::once,
     task::{Context, Poll},
-    time::{Duration, SystemTime, UNIX_EPOCH},
+    time::{SystemTime, UNIX_EPOCH},
 };
 use tower::{Layer, Service};
 use tower_http::sensitive_headers::{SetSensitiveRequestHeaders, SetSensitiveRequestHeadersLayer};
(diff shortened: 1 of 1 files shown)


ferranbt (approved):

0xOsiris (approved):

src/client/auth.rs
```diff
@@ -75,9 +75,10 @@ pub fn secret_to_bearer_header(secret: &JwtSecret) -> HeaderValue {
...
         secret
             .encode(&Claims {
-                iat: (SystemTime::now().duration_since(UNIX_EPOCH).unwrap()
-                    + Duration::from_secs(60))
-                .as_secs(),
+                iat: SystemTime::now()
+                    .duration_since(UNIX_EPOCH)
+                    .unwrap()
+                    .as_secs(),
                 exp: None,
```
0x416e746f6e, line 82: **suggestion:**

also shouldn't encode `exp` claim.  it's not in the spec, `op-node` doesn't send it either.  can save some bytes of the auth header.

src/client/auth.rs
```diff
@@ -75,9 +75,10 @@ pub fn secret_to_bearer_header(secret: &JwtSecret) -> HeaderValue {
         "Bearer {}",
         secret
             .encode(&Claims {
-                iat: (SystemTime::now().duration_since(UNIX_EPOCH).unwrap()
-                    + Duration::from_secs(60))
```
0x416e746f6e, line 79: **praise:**

thanks for this.  just finished troubleshooting an issue when due to slight drift b/w `rollup-boost` node and the builder, the builder was actually rejecting the engine API calls b/c the JWT was _more_ than 60 into the future.

PR #218, opened 2025-05-14
chore: refactor get_payload() function

karankurbur:
Refactor get_payload() by splitting into explicit functions for each step

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -577,82 +577,143 @@ impl RollupBoostServer {
         payload_id: PayloadId,
         version: Version,
     ) -> RpcResult<OpExecutionPayloadEnvelope> {
+        // Start fetching L2 payload early (we'll always need it)
         let l2_fut = self.l2_client.get_payload(payload_id, version);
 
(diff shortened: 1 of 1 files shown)


src/server.rs
```diff
@@ -577,82 +577,143 @@ impl RollupBoostServer {
         payload_id: PayloadId,
         version: Version,
     ) -> RpcResult<OpExecutionPayloadEnvelope> {
+        // Start fetching L2 payload early (we'll always need it)
```
Dzejkop, line 580: Unless a tokio task is spawned under the hood this future will only begin resolving once it's `.await`ed on. Consider wrapping it in a `tokio::spawn` to start resolving immediately

0xForerunner: It's being joined below though, so that should be fine.

0xForerunner (commented):
Not sure this PR is necessary. Original impl look a bit better to my eye.

src/server.rs
Dzejkop: true, but then there's no difference between this and spawning this future at the "join site". Plus the comment is misleading

PR #201, opened 2025-05-01
feat: process Engine API requests sequentially

0xKitsune:
Addresses #188, simplifies #192

This PR updates the ProxyService to process all Engine API requests sequentially.  Instead of forwarding Engine requests directly to the inner service, they’re now enqueued along with a response channel, and a separate task processes the queue. This guarantees that Engine API calls are processed sequentially while still allowing all other RPC calls to be processed concurrently.

PR #221, opened 2025-05-15
Handle Inconsistent Response in Miner Api

0xForerunner:
Audit findings:

All requests which are to be forwarded should not be treated equally. Specifically, there are eth_ and miner_ requests. If for some reason miner_ succeeds on the L2-EL but fails on the builder (or vice versa), it can result in discrepancy with critical information for consensus, such as GasPrice and GasLimit. This can lead to a block created by the builder with a higher gas limit than what L2-EL expects. As a result, the L2-EL will mark this block invalid and potentially flag all the future builder blocks as invalid invalid as well, effectively disabling builder functionality. 

Recommendation(s): Increase the robustness of the forwarding mechanism. The miner_ should be forwarder to the builder only if the call to EL-L2 succeeds. If the builder fails to process the request, the failure should be logged. In addition, the system should set the execution_status to Disabled to prevent further issues.


0xForerunner:
I've set execution mode to fallback here, although I'm not convinced this is the best way to handle this situation. I think setting the health to partial content is probably sufficient. Let me know what you think!

0xForerunner:
Since #201 was closed I cherry-picked this back onto main.

src/proxy.rs
```diff
@@ -131,8 +149,32 @@ where
...
+                let (builder_res, l2_res) = tokio::join!(
+                    builder_client.forward(builder_req, builder_method),
+                    service.l2_client.forward(l2_req, method)
+                );
+                if builder_res.is_ok() != l2_res.is_ok() {
+                    error!(target: "proxy::call", message = "inconsistent miner api responses from builder and L2");
+                    let mut execution_mode = service.execution_mode.lock();
+                    if *execution_mode == ExecutionMode::Enabled {
+                        *execution_mode = ExecutionMode::Disabled;
+                        // Drop before aquiring health lock
```
0xOsiris, line 171: I'm not sure if it's best to touch the `ExecutionMode` here. The health status change here will already trigger op-conductor to switch the active sequencer.

src/proxy.rs
```diff
@@ -1,20 +1,28 @@
...
 /// Requests that should be forwarded to both the builder and default execution client
-const FORWARD_REQUESTS: [&str; 6] = [
+const FORWARD_REQUESTS: [&str; 2] = [
     "eth_sendRawTransaction",
     "eth_sendRawTransactionConditional",
+];
+
+/// Handle these similar to FORWARD_REQUESTS, but enforce consistant responses
+/// between the builder and default execution client.
+const MINER_REQUESTS: [&str; 4] = [
```
0xOsiris, line 25: This issue is only relevant to `miner_setMaxDASize` as it's the only API used by OP. We could remove all other miner API methods here entirely. op-reth doesn't expose the API for anything else

src/proxy.rs
0xOsiris: Ideally the health status here sticks until there is no discrepancy in DA limits between the sequencer, and the builder

src/proxy.rs
```diff
@@ -131,8 +149,32 @@ where
...
+                    HttpRequest::from_parts(parts.clone(), HttpBody::from(body_bytes.clone()));
+                let builder_method = method.clone();
+                let mut builder_client = service.builder_client.clone();
+
+                let l2_req = HttpRequest::from_parts(parts, HttpBody::from(body_bytes));
+                let (builder_res, l2_res) = tokio::join!(
+                    builder_client.forward(builder_req, builder_method),
+                    service.l2_client.forward(l2_req, method)
+                );
+                if builder_res.is_ok() != l2_res.is_ok() {
```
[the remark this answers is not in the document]
0xForerunner: No because we can return that error to the caller, and they can handle that as they wish.

src/proxy.rs
0xForerunner: Ah good catch, will update!

ferranbt:
I will check more in detail next week but can we add a unit test to check this behaviour?

src/proxy.rs
0xForerunner: I think letting health status revert should be okay, as long as a new leader election is triggered.

0xForerunner:
@ferranbt Yeah will definitely add tests. Was just waiting to get settled on the solution. Just pushed a couple commits and I think this is currently as good as we can do. Added a retry layer into our HTTP client which should hopefully prevent ever running into this situation.

src/consistent_request.rs
```diff
@@ -0,0 +1,222 @@
...
+            builder_client,
+            has_disabled_execution_mode,
+            req_tx,
+            res_rx,
+            probes,
+            execution_mode,
+        };
+
+        let clone = manager.clone();
+        tokio::spawn(async move {
```
Dzejkop, line 60: Consider extracting this future to a separate async function - something like `async fn handle_requests`

src/consistent_request.rs
```diff
@@ -0,0 +1,222 @@
...
+    ) -> eyre::Result<()> {
+        // We send the l2 request first, because we need to avoid the situation where the
+        // l2 fails and the builder succeeds. If this were to happen, it would be too dangerous to
+        // return the l2 error response back to the caller, since the builder would now have an
+        // invalid state. We can return early if the l2 request fails.
+        let mut manager = self.clone();
+        let parts_clone = parts.clone();
+        let body_clone = body.clone();
+        let res_sender_clone = res_tx.clone();
+        tokio::spawn(async move {
```
Dzejkop, line 110: the result of this future is immediately awaited on, why are we spawning a task here?

src/consistent_request.rs
```diff
@@ -0,0 +1,222 @@
...
+                .send_to_l2(parts_clone, body_clone, res_sender_clone)
+                .await
+        })
+        .await??;
+
+        loop {
+            let mut manager = self.clone();
+            let parts_clone = parts.clone();
+            let body_clone = body.clone();
+            match tokio::spawn(
```
Dzejkop, line 121: Same as https://github.com/flashbots/rollup-boost/pull/221/files#r2102515472

src/consistent_request.rs
Dzejkop: If that's the case maybe we should drop a comment explaining it? Or consider an explicit cancellation token

tests/miner_set_max_da_size.rs
avalonche, line 1: are there meant to be any assertions on this test? It would be nice to have it test the case where the builder fails to respond to the call but happy to have it as a follow up

src/consistent_request.rs
```diff
@@ -0,0 +1,226 @@
...
+        builder_client: HttpClient,
+        probes: Arc<Probes>,
+        execution_mode: Arc<Mutex<ExecutionMode>>,
+    ) -> Self {
+        let (req_tx, mut req_rx) = watch::channel(None);
+        let (res_tx, mut res_rx) = watch::channel(None);
+        req_rx.mark_unchanged();
+        res_rx.mark_unchanged();
+
+        let has_disabled_execution_mode = Arc::new(AtomicBool::new(false));
```
avalonche, line 46: why is this initially set to false? shouldn't it be derived from the execution mode?

src/proxy.rs
```diff
@@ -120,19 +145,27 @@ where
...
-                let builder_method = method.clone();
-                let mut builder_client = service.builder_client.clone();
-                tokio::spawn(async move {
-                    let _ = builder_client.forward(builder_req, builder_method).await;
-                });
-
-                let l2_req = HttpRequest::from_parts(parts, HttpBody::from(body_bytes));
-                info!(target: "proxy::call", message = "forward request to default execution client", ?method);
-                service.l2_client.forward(l2_req, method).await
+                if method == MINER_SET_MAX_DA_SIZE {
```
avalonche, line 148: would be useful here to add docs why `MINER_SET_MAX_DA_SIZE` is treated differently here

src/consistent_request.rs
```diff
@@ -0,0 +1,226 @@
...
+        match l2_res {
+            Ok(t) => {
+                let (parts, body) = t.into_parts();
+                let (body_bytes, _) =
+                    http_helpers::read_body(&parts.headers, body, u32::MAX).await?;
+                res_tx.send(Some(Ok((parts, body_bytes))))?;
+                Ok(())
+            }
+            Err(e) => {
+                let msg = format!("failed to send request to l2: {e}");
```
avalonche, line 154: should `self.probes.set_health` be set to unhealthy here?

src/consistent_request.rs
```diff
@@ -0,0 +1,226 @@
...
+                    .has_disabled_execution_mode
+                    .load(std::sync::atomic::Ordering::SeqCst)
+                {
+                    let mut execution_mode = self.execution_mode.lock();
+                    *execution_mode = ExecutionMode::Enabled;
+                    // Drop before aquiring health lock
+                    drop(execution_mode);
+                    self.has_disabled_execution_mode
+                        .store(false, std::sync::atomic::Ordering::SeqCst);
+                    info!(target: "proxy::call", message = "setting execution mode to Enabled");
```
avalonche, line 180: should `self.probes.set_health` be reset here if the call failed previously

src/consistent_request.rs
```diff
@@ -0,0 +1,226 @@
...
+                let (parts, body) = t.into_parts();
+                let (body_bytes, _) =
+                    http_helpers::read_body(&parts.headers, body, u32::MAX).await?;
+                res_tx.send(Some(Ok((parts, body_bytes))))?;
+                Ok(())
+            }
+            Err(e) => {
+                let msg = format!("failed to send request to l2: {e}");
+                res_tx.send(Some(Err(e)))?;
+                Err(eyre!(msg))
```
avalonche, line 156: error logging would be useful here

src/consistent_request.rs
```diff
@@ -0,0 +1,226 @@
...
+            let parts_clone = parts.clone();
+            let body_clone = body.clone();
+            match tokio::spawn(
+                async move { manager.send_to_builder(parts_clone, body_clone).await },
+            )
+            .await?
+            {
+                Ok(_) => return Ok(()),
+                Err(_) => {
+                    tokio::time::sleep(Duration::from_millis(200)).await;
```
avalonche, line 129: error logging would be useful here

tests/miner_set_max_da_size.rs
```diff
@@ -0,0 +1,15 @@
...
+mod common;
+
+#[tokio::test]
+async fn miner_set_max_da_size() -> eyre::Result<()> {
+    let harness = RollupBoostTestHarnessBuilder::new("miner set max da size")
+        .with_isthmus_block(5)
+        .build()
+        .await?;
+    let mut block_generator = harness.block_generator().await?;
+    block_generator.generate_builder_blocks(10).await?;
```
ferranbt, line 12: Is this test finished?

src/consistent_request.rs
```diff
@@ -0,0 +1,222 @@
...
+        res_tx: watch::Sender<Option<Result<(response::Parts, Vec<u8>), BoxError>>>,
+    ) -> eyre::Result<()> {
+        // We send the l2 request first, because we need to avoid the situation where the
+        // l2 fails and the builder succeeds. If this were to happen, it would be too dangerous to
+        // return the l2 error response back to the caller, since the builder would now have an
+        // invalid state. We can return early if the l2 request fails.
+        let mut manager = self.clone();
+        let parts_clone = parts.clone();
+        let body_clone = body.clone();
+        let res_sender_clone = res_tx.clone();
```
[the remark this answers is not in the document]
0xForerunner: No you need to clone for the loop

tests/miner_set_max_da_size.rs
0xForerunner: Test is incomplete haha. Will complete shortly

src/proxy.rs
0xForerunner: Done!

src/consistent_request.rs
0xForerunner: The health should be automatically reset when the builder successfully returns the next payload.

tests/miner_set_max_da_size.rs
0xForerunner: It is not haha. Working on that next!

tests/miner_set_max_da_size.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+impl ProxyHandler for MaxDaSizeHandler {
+    fn handle(
+        &self,
+        method: String,
+        params: Value,
+        _result: Value,
+    ) -> Pin<Box<dyn Future<Output = Option<Value>> + Send>> {
+        println!("method: {:?}", method);
```
avalonche, line 27:
```suggestion
```

avalonche (approved):

tests/miner_set_max_da_size.rs
```diff
@@ -0,0 +1,80 @@
...
+    let second_val = 2000000;
+    let _res = engine_api.set_max_da_size(second_val, second_val).await;
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    let found = handler.get_found_max_da_value();
+    assert_eq!(found, 0);
+
+    // restart the builder
+    client.unpause_container(harness.builder.id()).await?;
+
+    // let _res = engine_api.set_max_da_size(val, val).await?;
```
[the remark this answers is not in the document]
0xForerunner: Fixed, TY!

PR #233, opened 2025-05-19
Restructure the repo as a workspace

### LINKED ISSUE #223: Restructure the repository as a workspace
Ref https://github.com/flashbots/rollup-boost/pull/203#pullrequestreview-2822933116

ferranbt:
Closes #223

I have not changed any internal Rollup-boost logic but on the way we run the integration tests:
TODO


crates/rollup-boost/src/cli.rs
```diff
@@ -90,9 +90,7 @@ pub struct Args {
 
 impl Args {
     pub async fn run(self) -> eyre::Result<()> {
-        rustls::crypto::ring::default_provider()
-            .install_default()
-            .expect("Failed to install TLS ring CryptoProvider");
+        let _ = rustls::crypto::ring::default_provider().install_default();
```
ferranbt, line 93: Same as with the other changes. It fails if you try to install twice.

crates/rollup-boost/src/tests/common/mod.rs
```diff
@@ -242,7 +244,7 @@ impl RollupBoostTestHarnessBuilder {
         let timestamp = dt.format(&format)?;
 
         let dir = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-            .join("integration_logs")
+            .join("../../integration_logs")
```
ferranbt, line 245: I like to have the logs on the root folder.

crates/rollup-boost/src/tests/common/services/rollup_boost.rs
```diff
@@ -42,8 +45,6 @@ impl Default for RollupBoostConfig {
             &format!("--l2-jwt-path={}/jwt_secret.hex", *TEST_DATA),
             &format!("--builder-jwt-path={}/jwt_secret.hex", *TEST_DATA),
             "--log-level=trace",
-            "--tracing",
```
avalonche, line 45: why are tracing / metrics args removed?

crates/rollup-boost/src/cli.rs
avalonche: would have a warning log here instead of a silent failure

avalonche (approved):

0xOsiris (changes requested):
What do you think about moving all dependencies into the workspace root, and then reference workspace dependencies within the individual crates? I think this will speed up workspace compilation times with many shared dependencies across the two crates.

ferranbt:
> What do you think about moving all dependencies into the workspace root, and then reference workspace dependencies within the individual crates? I think this will speed up workspace compilation times with many shared dependencies across the two crates.

True! I only thought about the alloy/reth dependencies. I will move the shared ones to the workspace Cargo.toml

avalonche:
does this change how rollup-boost is run within the workspace? please also update the command line docs if so

PR #236, opened 2025-05-19
Add block builder tee proofs

fnerdman:
Adds a section on how flashtestations will be used as block proofs. not sure if we should add. Pros: adds to the bigger picture, implementation is easy once we have flashtestations, so we could add it to the current cycle. Cons: somewhat diverts from the path of trying to present a spec that can universally and transparently track attestations and their TTL.
 Don't merge for now, needs feedback.

### DIFF
--- a/specs/flashtestations.md
+++ b/specs/flashtestations.md
@@ -670,3 +670,96 @@ To establish trust in expected measurements, the TEE block builder must be built
 4. **Verification**: Independent parties can follow the build process and verify that it produces the same measurements
 
 This allows anyone to verify that the expected measurements correspond to the published source code.
+
+## Block Builder TEE Proofs
+
+The Flashtestations protocol can be extended to provide cryptographic guarantees that blocks were constructed by an authorized TEE-based block builder. This section describes how block builders running in a TEE can prove block authenticity through an onchain verification mechanism.
+
+### Core Mechanism
+
+The block builder TEE proof system works through a final transaction appended to each block. This transaction:
+
+1. Calls a designated smart contract method that accepts a block content hash
+2. Verifies the caller's authorization using `isAllowedPolicy(block_builder_policy_id, msg.sender)`
+3. Provides cryptographic evidence that the block was constructed by a valid TEE-based block builder
+
+The key insight is that the required private key to sign this transaction is protected within the TEE environment. Thus, only a genuine TEE-based block builder with the proper attestation can successfully execute this transaction.
+
+### Block Building Process
+
+When building a block, the TEE block builder:
+
+1. Produces a block according to the L2 protocol rules
+2. Computes the block content hash using the `ComputeBlockContentHash` function:
+
+```solidity
+function ComputeBlockContentHash(block, transactions) {
+    // Create ordered list of all transaction hashes
+    transactionHashes = []
+    for each tx in transactions:
+        txHash = keccak256(rlp_encode(tx))
+        transactionHashes.append(txHash)
+    
+    // Compute a single hash over block data and transaction hashes
+    // This ensures the hash covers the exact transaction set and order
+    return keccak256(abi.encode(
+        block.parentHash,
+        block.number,
+        block.timestamp,
+        transactionHashes
+    ))
+}
+```
+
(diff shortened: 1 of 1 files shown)


dmarzzz:
let's close this for now imo @fnerdman , when @avalonche starts implementing we can add this in

specs/flashtestations.md
```diff
@@ -670,3 +670,96 @@ To establish trust in expected measurements, the TEE block builder must be built
...
+    require(
+        IAllowlist(ALLOWLIST_CONTRACT).isAllowedPolicy(BLOCK_BUILDER_POLICY_ID, msg.sender),
+        "Unauthorized block builder"
+    );
+    
+    // At this point, we know:
+    // 1. The caller is a registered address from an attested TEE
+    // 2. The TEE is running an approved block builder workload (via policy)
+    
+    // Additional validation can be performed on the blockContentHash
```
Melvillian, line 739: What validation would this be? I can't think of anything useful we can do with a nonstandard hash like the one we have here. Solidity doesn't have retrospection so we can't look at the previous transactions to validate the blockContentHash onchain; all we can do is rely on the TEE to correctly compute this hash.

If we can't think of anything, can we remove this comment so it's not ambiguous what we mean here? That would simplify things for implementers

specs/flashtestations.md
```diff
@@ -670,3 +670,96 @@ To establish trust in expected measurements, the TEE block builder must be built
...
+- Enables signature verification without requiring state root dependencies
+- Supports future L1 verification of block authenticity across different rollup designs
+
+### Verification Contract
+
+The smart contract that verifies block builder TEE proofs would implement:
+
+```solidity
+// Block builder verification contract
+function verifyBlockBuilderProof(bytes32 blockContentHash) external {
```
Melvillian, line 728: I think we should add a `uint8 version` argument here, so we can easily communicate the version of flashtestations this verification is meant for. Without this for example, if we later decide that we want to we want to change how the blockContentHash is calculated, then we'll have no way to communicate to offchain verifiers how to recompute the `blockContentHash` offchain.

The downside of this is that it will slightly increase L1 gas cost because of the additional calldata. But it's just a single 8-bit value, so it's very little.

specs/flashtestations.md
```diff
@@ -670,3 +670,96 @@ To establish trust in expected measurements, the TEE block builder must be built
...
+        "Unauthorized block builder"
+    );
+    
+    // At this point, we know:
+    // 1. The caller is a registered address from an attested TEE
+    // 2. The TEE is running an approved block builder workload (via policy)
+    
+    // Additional validation can be performed on the blockContentHash
+    
+    emit BlockBuilderProofVerified(
```
Melvillian, line 741: we should add version here (see above comment)

specs/flashtestations.md
```diff
@@ -670,3 +670,96 @@ To establish trust in expected measurements, the TEE block builder must be built
...
+2. **Non-Transferability**: The proof cannot be stolen or reused by unauthorized parties due to TEE protection of signing keys
+3. **Policy Flexibility**: The system can adapt to new block builder implementations by updating the policy without contract changes
+4. **Auditability**: All proofs are recorded onchain for transparency and verification
+
+### Integration with Rollup Systems
+
+For rollup systems, this proof mechanism:
+
+1. Can be included as the final transaction in each block
+2. Enables L1 contracts to verify that blocks were built by authorized TEEs
```
Melvillian, line 763: Is this possible? L1 contracts do not have access to the EIP-4844 blobs that (see [this explainer on EIP-4844](https://www.eip4844.com/) specifically the wording "Blobs are persisted in beacon nodes, not in execution layer").

If we want to expose this data to L1's, we'd need some way to bridge this transaction call to the L1. Maybe we want to do that, but it'll increase the gas cost of this (which we want to keep as cheap as possible).

I think we should remove this line for now, as it's not clear it's possible.

Melvillian (commented):
High level looks good 👍 but I have some comments on specific parts.

fnerdman:
closing as out of scope for now

PR #237, opened 2025-05-20
Flashblocks rebase

avalonche:


src/tracing.rs
```diff
@@ -58,6 +58,34 @@ impl SpanProcessor for MetricsSpanProcessor {
             ])
             .collect::<Vec<_>>();
 
+        // 0 = no difference in gas build via builder vs l2
```
ferranbt, line 61: What is this for?

src/flashblocks/outbound.rs
```diff
@@ -0,0 +1,77 @@
...
+            let message = match serde_json::to_string(&payload) {
+                Ok(msg) => Message::Text(msg.into()),
+                Err(e) => {
+                    error!("Failed to serialize payload: {}", e);
+                    continue;
+                }
+            };
+
+            let mut clients = self.clients.lock().await;
+            clients.retain_mut(|sink| {
```
SozinM, line 63: Have we tested this part with 100/1000 clients, will it have problems because of blocking?

PR #239, opened 2025-05-20
[Draft] Add Engine API abstraction

ferranbt:
This PR adds an engine API trait to the RollupBoostServer such that there can be several implementations for the backend RB logic. Some of the names are TBD.


ferranbt:
Closing in favour of #270

PR #245, opened 2025-05-22
Call execution_mode only once in get_payload

ferranbt:


avalonche (approved):

src/server.rs
```diff
@@ -654,7 +655,7 @@ impl RollupBoostServer {
             self.probes.set_health(Health::Healthy);
 
             if let Ok(Some(payload)) = builder_payload {
-                if self.execution_mode().is_dry_run() {
+                if execution_mode.is_dry_run() {
```
0xKitsune, line 658: There is an edge case here where execution mode is enabled when this function is invoked but later switched to `DryRun` before receiving the payloads. In this case, the builder block will be selected. If this is an acceptable edge case then lgtm.

0xKitsune (approved):

PR #262, opened 2025-05-28
feat: retry `miner_setMaxDASize` calls to builder

0xKitsune:
This PR adds support for retrying `miner_setMaxDASize` requests in the Rollup Boost server.

When a request is received, it first attempts to call the L2 client. If that call fails, the response is returned immediately without retry. If the L2 call succeeds and the builder call fails, the builder call is retried on a fixed interval.

If execution mode is enabled when the builder failure occurs, execution mode is updated `Disabled` and the health probe is marked as `PartialContent`. Once the retry eventually succeeds, execution mode is re-enabled only if it was previously disabled by this logic. This ensures that manual operator overrides are respected and not unintentionally reverted.


0xKitsune:
Closing with preference for #302

0xKitsune:
Closing with preference for #302

PR #263, opened 2025-05-28
Add builder playground to CI

avalonche:


.github/workflows/builder-playground.yml
```diff
@@ -0,0 +1,70 @@
...
+          sudo apt-get install -y libsqlite3-dev libfontconfig1-dev libfontconfig
+          cargo install --git https://github.com/flashbots/contender --bin contender --force
+
+      - name: Install op-rbuilder
+        run: |
+          cargo install --git https://github.com/flashbots/op-rbuilder --bin op-rbuilder --force
+
+      - name: Setup Builder Playground
+        shell: bash
+        run: |
```
ferranbt, line 56: You should be able to install playground from here https://github.com/flashbots/flashbots-toolchain

.github/workflows/builder-playground.yml
```diff
@@ -0,0 +1,70 @@
...
+            go get github.com/flashbots/builder-playground
+            go install github.com/flashbots/builder-playground 
+  
+      - name: Run Builder Playground
+        shell: bash
+        run: |
+            op-rbuilder -- node --builder.playground
+            builder-playground cook opstack --external-builder http://host.docker.internal:4444
```
ferranbt, line 64: If you run external-builder=op-reth works as well unless you explicitely to run the external builder outside.

PR #270, opened 2025-05-29
Flashblocks with extension trait

ferranbt:


karim-agha (approved):

ferranbt:
>some of the README and .env.example updates are missing. is this intentional?

Forgot to copy them over. Will do.

crates/rollup-boost/src/flashblocks/launcher.rs
```diff
@@ -0,0 +1,28 @@
...
+        let (tx, rx) = mpsc::channel(100);
+
+        let receiver = FlashblocksReceiverService::new(flashblocks_url, tx)?;
+        tokio::spawn(async move {
+            let _ = receiver.run().await;
+        });
+
+        let service = FlashblocksService::new(builder_url, outbound_addr)?;
+        let mut service_handle = service.clone();
+        tokio::spawn(async move {
```
0xOsiris, line 22: Rather than spawning two green threads here, and dropping the `JoinHandle` to both. Consider using `tokio::select!` on both the futures with a cancellation token to cancel the two branches if either resolves, or an interrupt occurs.

crates/rollup-boost/src/flashblocks/args.rs
```diff
@@ -0,0 +1,20 @@
+use clap::Parser;
+
+#[derive(Parser, Clone, Debug)]
+pub struct FlashblocksArgs {
+    /// Enable Flashblocks client
+    #[arg(long, env, default_value = "false")]
+    pub flashblocks: bool,
+
+    /// Flashblocks Builder WebSocket URL
+    #[arg(long, env, default_value = "ws://localhost:1111")]
```
0xOsiris, line 10:
```suggestion
    #[arg(long, env, default_value = "ws://127.0.0.1:8546")]
```

crates/rollup-boost/src/flashblocks/args.rs
```diff
@@ -0,0 +1,20 @@
...
+    pub flashblocks_builder_url: String,
```
0xOsiris, line 11: Consider using type `Url` here to avoid an additional type conversion

crates/rollup-boost/src/flashblocks/inbound.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub async fn run(self) {
+        loop {
+            match self.connect_and_handle().await {
+                Ok(()) => break,
+                Err(e) => {
+                    error!(
+                        message = "Flashblocks receiver connection error, retrying in 5 seconds",
+                        error = %e
+                    );
+                    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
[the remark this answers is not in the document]
ferranbt: If the remote builder, I prefer not to have it pinging so often. 5 seconds seems a good enough tradeoff.

crates/rollup-boost/src/flashblocks/inbound.rs
```diff
@@ -0,0 +1,58 @@
...
+                }
```
0xKitsune, line 31:
```suggestion
        while let Err(e) = self.connect_and_handle().await {
            error!(
                "Flashblocks receiver connection error, retrying in 5 seconds: {e}"
            );
            tokio::time::sleep(std::time::Duration::from_secs(5)).await;
        }  
```

crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -0,0 +1,318 @@
...
+
+        // Check base payload rules
+        match (payload.index, payload.base) {
+            // First payload must have a base
+            (0, None) => return Err(FlashblocksError::MissingBasePayload),
+            (0, Some(base)) => self.base = Some(base),
+            // Subsequent payloads must have no base
+            (_, Some(_)) => return Err(FlashblocksError::UnexpectedBasePayload),
+            (_, None) => {} // Non-zero index without base is fine
+        }
```
0xKitsune, line 86: Nit: consider simplifying to something like the following:

```suggestion
if payload.index == 0 {
    // First payload must include a base
    if let Some(base) = payload.base {
        self.base = Some(base);
    } else {
        return Err(FlashblocksError::MissingBasePayload);
    }
} else if payload.base.is_some() {
    // Non-zero index must not include a base
    return Err(FlashblocksError::UnexpectedBasePayload);
}
```

crates/rollup-boost/src/flashblocks/inbound.rs
0xKitsune, line 31: Nit: consider simplifying to
```suggestion
        if let Err(e) = self.connect_and_handle().await {
            error!("Flashblocks receiver connection error, retrying in 5 seconds: {e}");
            tokio::time::sleep(Duration::from_secs(5)).await;
        } else {
            break;
        }
```

crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -0,0 +1,318 @@
...
+            let event = stream.recv().await;
+            match event {
+                Some(event) => {
+                    self.on_event(FlashblocksEngineMessage::FlashblocksPayloadV1(event))
+                        .await
+                }
+                None => break,
+            }
+        }
+    }
```
0xKitsune, line 267: Consider simplifying to

```suggestion
 while let Some(event) = stream.recv().await {
        self.on_event(FlashblocksEngineMessage::FlashblocksPayloadV1(event)).await;
    }
```

avalonche (approved):

PR #271, opened 2025-05-29
Release nightly docker images for Rollup-boost

### LINKED ISSUE #164: Create nightly docker images
Add nightly docker images for Rollup-boost

ferranbt:
Closes #164

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ on:
       - main
   schedule:
     # Run nightly at 2 AM UTC
-    - cron: "0 2 * * *"
+    - cron: "0 1 * * *"
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1,10 +1,13 @@
-name: Release
+name: Build Docker Image
```
ferranbt, line 1: I also renamed this workflow because it had the same as the release.yaml workflow and it was confusing for manual dispatch.

avalonche (approved):

avalonche:
why do we need a nightly image when a new image is built for every commit?

ferranbt:
It is useful to have a rollup_boost:nightly to point to

Issue #253, opened 2025-05-23
Add a label to signal whether the block built was deterministic


ferranbt:
related #252

PR #274, opened 2025-05-29
Add builder has payload attribute to metrics

avalonche:
Resolves https://github.com/flashbots/rollup-boost/issues/253. Removed the `has_attributes` since it was not being included in the span attributes and unused


ferranbt:
Was this for #253?

ferranbt (approved):

PR #275, opened 2025-05-30
Update reth to 1.4.3

avalonche:


avalonche:
@0xForerunner I removed the `invalid_jwt` test as rollup-boost never validated the jwt secret in the first place. would you mind taking a look to see if this behaviour is valid?

PR #276, opened 2025-05-30
Bump reth to v1.4.4

therustmonk:
Updated dependencies to the latest reth 1.4.4. I see that #275 already exists, but it's based on the `feat/fb-for-real` branch, whereas this one targets `main`.

Let me know what’s the best course of action here, but this branch is currently needed to open a PR in op-rbuilder, which requires the latest version of reth (1.4.4).

cc @ferranbt @avalonche @SozinM 

UPD: Actually, the main branch is missing quite a lot for update, and now I've essentially made changes on top of the PR #275. Not sure should I use `feat/fb-for-real` or `main`.


ferranbt:
Closing in favour of #278

PR #277, opened 2025-06-03
Do not update health status when execution mode = Disabled

dpulitano:
On builder payload failures, we should only update the health status if the execution mode is Enabled

Instead, I added a WARN log

Testing still in progress...


avalonche (approved):

dpulitano:
will be handled by #295 instead

PR #278, opened 2025-06-03
Upgrade reth to 1.4.7

avalonche:


crates/rollup-boost/Cargo.toml
```diff
@@ -24,12 +24,14 @@ alloy-serde.workspace = true
 
 tokio-tungstenite.workspace = true
 
-jsonrpsee = { version = "0.24", features = ["server", "http-client", "macros"] }
+# TODO: update to latest release when it is published
+# jsonrpsee = { version = "0.25.1", features = ["server", "http-client", "macros"] }
+jsonrpsee = { git = "https://github.com/paritytech/jsonrpsee", rev = "f04afa740e55db60dce20d9839758792f035ffff", features = ["server", "http-client", "macros"]}
```
ferranbt, line 29: I believe this is because reth update?

ferranbt: Reth pins the version 0.25.1, can't we use that version?

PR #294, opened 2025-06-05
fix: use CLI configured timeouts for proxy Http clients

0xOsiris:


crates/rollup-boost/src/client/http.rs
```diff
@@ -46,7 +46,7 @@ impl HttpClient {
         let client = Client::builder(TokioExecutor::new()).build(connector);
 
         let client = ServiceBuilder::new()
-            .layer(TimeoutLayer::new(Duration::from_secs(1)))
+            .layer(TimeoutLayer::new(Duration::from_secs(timeout)))
```
0xKitsune, line 49: The `RpcClient` timeout is initialized with `Duration::from_millis`. We should update this to match.

0xOsiris: Ah nice, good catch!

0xKitsune (approved):

Issue #292, opened 2025-06-05
Health status affected by builder when execution mode is disabled


0xKitsune:
Rollup Boost's health check logic polls the builder for the latest unsafe block and marks the service as `PartialContent` if the builder is out of sync or unreachable. This behavior makes sense when `execution_mode=enabled` as the builder is actively used for block production and failover ensures payloads continue to be served.

```rust
impl HealthHandle {
    /// Periodically checks that the latest unsafe block timestamp is not older than the
    /// the current time minus the max_unsafe_interval.
    pub fn spawn(self) -> JoinHandle<()> {
        // --snip--
            loop {
                let latest_unsafe = match self
                    .builder_client
                    .get_block_by_number(BlockNumberOrTag::Latest, false)
                    .await
                {
                    Ok(block) => block,
                    Err(e) => {
                        warn!(target: "rollup_boost::health", "Failed to get unsafe block from builder client: {} - updating health status", e);
                        self.probes.set_health(Health::PartialContent);
                        sleep_until(Instant::now() + self.health_check_interval).await;
                        continue;
                    }
                };
        // --snip--
}
```

However, in `execution_mode=disabled`, Rollup Boost ignores the builder for all payload related logic and acts as a "vanilla" OP Stack sequencer. Despite this, the builder sync status is still polled and health is downgraded to `PartialContent` if the builder is not synced causing `op-conductor` to fail over, which may not be the desired behavior for some chain operators.

Opening this issue to clarify if the builder's health should influence Rollup Boost's health status when `execution_mode=disabled`?

ferranbt:
Is this fixed as part of dpulitano:danp/disabled-mode-health-fix?

dpulitano:
My main concern when running OP Stack chains is having a cohesive way for op-conductor to automatically know not to failover whenever the active sequencer has execution_mode = disabled, since the purpose of execution_mode=disabled is to solely rely on the L2 execution client.

My first preference would be to update the healthcheck endpoint to only consider L2 health when execution mode is disabled (return 200 if the L2 is healthy, even if the builder is unhealthy). A Prom metric could show if the builder is healthy or not and can be monitored separately from the healthcheck endpoint

Another option I'd be ok with but I feel is more work is to continue returning 206 if the L2 is healthy and Builder is unhealthy, but include content or a header in the response if the execution mode is disabled (e.g. `ExecutionMode = Disabled`). Then a PR is made for op-conductor to parse that and automatically know not to failover

[...]

PR #295, opened 2025-06-06
Osiris/update health handle

0xOsiris:
Addresses #292


crates/rollup-boost/src/health.rs
```diff
@@ -35,7 +54,9 @@ impl HealthHandle {
                     Ok(block) => block,
                     Err(e) => {
                         warn!(target: "rollup_boost::health", "Failed to get unsafe block from builder client: {} - updating health status", e);
-                        self.probes.set_health(Health::PartialContent);
+                        if !self.execution_mode.lock().is_disabled() {
```
dpulitano, line 57: I feel dry-run mode should work the same way as disabled since the goal there is also to just rely on L2 block-building while seeing how the builder client behaves when receiving engine calls

What do you think about having PartialContent being returned only when execution_mode=enabled?

crates/rollup-boost/src/health.rs
```diff
@@ -308,6 +333,7 @@ mod tests {
 
         let health_handle = HealthHandle {
             probes: probes.clone(),
+            execution_mode: Arc::new(Mutex::new(ExecutionMode::Enabled)),
             builder_client: builder_client.clone(),
             health_check_interval: Duration::from_secs(60),
             max_unsafe_interval: 5,
```
dpulitano, line 339: Can we have another test for making sure the Health probe is 200 when ExecutionMode != Enabled and unsafe head lags more than the unsafe interval?

crates/rollup-boost/src/cli.rs
```diff
@@ -231,6 +232,9 @@ impl Args {
                 // The server has already shut down by itself
                 info!("Server stopped");
             }
+            _ = health_handle => {
+                info!("Health check task stopped");
```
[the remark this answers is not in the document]
0xOsiris: Well this will completely crash the program if this future resolves. So, maybe an `error` actually is more appropriate

crates/rollup-boost/src/cli.rs
```diff
@@ -39,7 +39,7 @@ pub struct Args {
     pub health_check_interval: u64,
 
     /// Max duration in seconds between the unsafe head block of the builder and the current time
-    #[arg(long, env, default_value = "5")]
+    #[arg(long, env, default_value = "20")]
```
dpulitano, line 42: I feel this number should be optimized for when execution mode is enabled. In that case, 20s seems like a lot (We usually consider a sequencer unhealthy after 10s of unsafe head lag)

Curious to hear your thoughts on this

0xOsiris: I'll update to 10 seconds. It seemed like 5 was consistently hitting on the active sequencer, so it seemed a little low

crates/rollup-boost/src/cli.rs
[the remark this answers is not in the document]
dpulitano: Ah understood. Fine with this then 👍

dpulitano (approved):
LGTM!

crates/rollup-boost/src/cli.rs
dpulitano: Cool. yeah hopefully that's a good sweet spot

0xKitsune (approved):

PR #300, opened 2025-06-06
Initial batch of metrics for flashblocks service

ferranbt:
This PR adds an initial batch of metrics for flashblocks.

### DIFF
--- a/crates/rollup-boost/src/flashblocks/inbound.rs
+++ b/crates/rollup-boost/src/flashblocks/inbound.rs
@@ -1,4 +1,5 @@
 use super::primitives::FlashblocksPayloadV1;
+use crate::flashblocks::metrics::FlashblocksWsInboundMetrics;
 use futures::StreamExt;
 use tokio::sync::mpsc;
 use tokio_tungstenite::{connect_async, tungstenite::Message};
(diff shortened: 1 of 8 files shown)


crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -222,6 +226,8 @@ impl FlashblocksService {
     async fn on_event(&mut self, event: FlashblocksEngineMessage) {
         match event {
             FlashblocksEngineMessage::FlashblocksPayloadV1(payload) => {
+                self.metrics.messages_processed.increment(1);
```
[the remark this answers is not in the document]
SozinM: Am'i wrong or you increase it here and when you receive a test message?

crates/rollup-boost/src/flashblocks/inbound.rs
```diff
@@ -17,13 +19,16 @@ impl FlashblocksReceiverService {
...
+            metrics: Default::default(),
         }
     }
 
     pub async fn run(self) {
         loop {
             if let Err(e) = self.connect_and_handle().await {
                 error!("Flashblocks receiver connection error, retrying in 5 seconds: {e}");
+                self.metrics.reconnect_attempts.increment(1);
+                self.metrics.connection_status.set(0);
```
avalonche, line 31: this should be part of the health status instead of metrics?

ferranbt: What do you mean? which health status?

Issue #288, opened 2025-06-05
Memory leak for latest rollup-boost on higher throughput chain


cody-wang-cb:
Been testing the latest versions of rollup-boost, it runs fine on devnet, but noticed some memory leak issues when runnning on Base Sepolia.

<img width="305" alt="Image" src="https://github.com/user-attachments/assets/12a8946a-9de9-4986-8089-e7c9b1109f70" />

It uses a lot more memory and the usage doesn't stop growing as the chain progresses.

I don't have `--flashblocks` flag on so it should be unrelated to any of the websocket related stuff. 

I don't seem to see anything obvious so just been cherry-picking commits and deploy to Base Sepolia.

So far tested that
- baebd910c33e9a821ba8640082d33a507a62dc46
- 6912f3fd4bbea46366d1d3dff9da9a4f5f684b71
all have memory issues

187c1fdf540722934564cb465f2e0f7ac7735748 does not have memory issues, so it's probably some changes in between that caused this

avalonche:
do you have metrics on the rpc requests being sent to rollup-boost? that PR is strictly related to the miner_ rpc calls

PR #302, opened 2025-06-06
fix: remove miner retry

0xKitsune:
This PR addresses #288 by removing retry logic for `miner_setMaxDASize`. Instead, we now treat it like any other forwarded request.

`op-batcher` sends `miner_setMaxDA` calls every 10 seconds regardless of the current throttle state, ensuring that the builder receives periodic updates even if individual requests fail. This makes retry logic unnecessary at the proxy layer.

Chain operators should rely on monitoring and alerting to ensure failures in `miner_setMaxDA` calls are caught. If preferred, we could also consider adding explicit error logging when forward requests fail.


crates/rollup-boost/src/proxy.rs
```diff
@@ -172,7 +143,12 @@ where
 
                 // Fire and forget the builder request
                 tokio::spawn(async move {
-                    let _ = builder_client.forward(buffered_clone, method_clone).await;
+                    if let Err(err) = builder_client
+                        .forward(buffered_clone, method_clone.clone())
+                        .await
+                    {
+                        tracing::error!(method = %method_clone, error = ?err, "Builder forward request failed");
```
avalonche, line 150: I believe the .forward method already has an error log, though it might be worth adding the method in that error log (atm it only logs the http code)

0xKitsune: Good catch, I removed the redundant logging. The method should already be logged via the `forward` method.

Issue #306, opened 2025-06-08
bug: race condition b/w flashblocks vs. fallback `getPayload` causes the sequencer to stall


0x416e746f6e:
tl;dr:  `rollup-boost` can return the same block for two different payload ids, which causes the sequencer to log `"Sequencer paused until new events"` and stall forever (only restart would revive it).

the scenario is most likely to occur during the periods when the sequencer is incrementally advancing safe head upon reading batcher's channel (see: https://github.com/ethereum-optimism/optimism/issues/15393)

what can happen is:

- `N`: when handling `getPayload` from the sequencer `rollup-boost`, for whatever reason, instead of returning the block that it had assembled from the flashblocks stream, falls back to the engine api, and issues the `getPayload` call to the builder
  - builder returns a valid block, and that's what is handed over to the sequencer
- `N+1`: then for the _next_ `getPayload` from the sequencer `rollup-boost` would send still the same block `N`, but this time it's the one it assembled from the flashblocks stream
  - sequencer receives a block with the hash it already knows, and the logic [here](https://github.com/ethereum-optimism/optimism/blob/v1.13.3/op-node/rollup/sequencing/sequencer.go#L455-L466) is not triggered, which blocks its event loop forever

#### jrpc calls on `rollup-boost` side:

![Image](https://github.com/user-attachments/assets/a6dc8de3-fb24-4edc-8a94-6b4ea0614f93)

note that for 2 consecutive payload ids `0x03af34acbbd4615d` and `0x03464ebe9ab36481` the same block `0x07ae1e557c92eb007c493ff13775b3ab950a328eefdb0e6de7fc68bbd1196278` is returned by `rollup-boost`

#### jrpc calls on builder side:

<img width="1155" alt="Image" src="https://github.com/user-attachments/assets/df609ad5-bf02-45aa-9b54-8c50d129f932" />

the response for payload id `0x03af34acbbd4615d` actually came from the builder

0x416e746f6e:
could be related to https://github.com/flashbots/rollup-boost/pull/149

PR #307, opened 2025-06-09
Fix FB race

SozinM:
Run make fmt
Added check to ensure that we returning block constructed from flashblock
Removed double lock - this changed behavior a bit, but even if into_envelope fails we would have FlashblockBuilder::new() in best_payload

### DIFF
--- a/crates/rollup-boost/src/flashblocks/service.rs
+++ b/crates/rollup-boost/src/flashblocks/service.rs
@@ -208,7 +208,7 @@ impl FlashblocksService {
         // Check that we have flashblocks for correct payload
         if *self.current_payload_id.read().await != payload_id {
             // We have outdated `current_payload_id` so we should fallback to get_payload
-            // Clearing best_payload in here would cause situation when old `get_payload` would clear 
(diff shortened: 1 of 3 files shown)


crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -203,13 +203,19 @@ impl FlashblocksService {
     pub async fn get_best_payload(
         &self,
         version: PayloadVersion,
+        payload_id: PayloadId,
     ) -> Result<Option<OpExecutionPayloadEnvelope>, FlashblocksError> {
+        // Check that we have flashblocks for correct payload
+        if *self.current_payload_id.read().await != payload_id {
```
ferranbt, line 209: I like it as a safe check, just noting that this would mean that the flashblocks module missed a FCU call to reset its payload id.

SozinM: I also added best_payload cleaning on FCU with attr, otherwise it won't be cleaned

SozinM:
Fixes https://github.com/flashbots/rollup-boost/issues/306
@0x416e746f6e

PR #308, opened 2025-06-09
Use User-agent in RB header

ferranbt:
Uses `User-Agent` key in Rollup-boost Http call header that includes the version.


0x416e746f6e:
this is what the header sends:

```
0.1.0-VERGEN_
```

^-- we need something better

PR #309, opened 2025-06-09
ws proxy: brotli compress each message from upstream before sending downstream

haardikk21:
The Websocket Proxy serves many downstream clients, and the raw JSON payload of the Flashblock stream adds up pretty quickly. Estimated monthly egress is quite high.

This PR introduces Brotli compression on the proxy level, giving us an average improvement of ~5.5-6x on the payload size tested against Base Sepolia. On Base Mainnet where blocks are larger, the difference is likely going to be even larger. 

Downstream clients (like Flashblocks aware preconf nodes) and direct WS consumers will need to implement decompression of the message before parsing it as JSON.

### DIFF
--- a/crates/websocket-proxy/src/main.rs
+++ b/crates/websocket-proxy/src/main.rs
@@ -222,7 +222,7 @@ async fn main() {
             {
                 let mut compressor =
                     brotli::CompressorWriter::new(&mut compressed_data_bytes, 4096, 5, 22);
-                compressor.write_all(&data_bytes).unwrap();
+                compressor.write_all(data_bytes).unwrap();
             }
             compressed_data_bytes
         } else {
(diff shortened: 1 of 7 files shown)


crates/websocket-proxy/src/main.rs
```diff
@@ -18,6 +18,8 @@ use clap::Parser;
 use dotenvy::dotenv;
 use metrics_exporter_prometheus::PrometheusBuilder;
 use rate_limit::RedisRateLimit;
+#[cfg(feature = "compression")]
```
danyalprout, line 21: @haardikk21 what do you think about using an arg for this? (can add [here](https://github.com/flashbots/rollup-boost/blob/main/crates/websocket-proxy/src/main.rs#L31)).

Feels like a bit overkill to add a new compile feature for this, I can imagine this is something we may want to toggle quickly on a devnet etc without having to recompile

ferranbt (approved):

haardikk21:
> Do you see a need to do this on Rollup-boost <> builder side?

We can discuss it. Our main goal right now was to try to reduce egress size/costs and this is the main thing that contributes to it. It isn't an immediate concern to do this for the internal communication.

Issue #304, opened 2025-06-06
improv (flashblocks):  implement ping/pong on websocket connection + set deadline on reads


0x416e746f6e:
- presently the flashblocks websocket connection is one-way (`rollup-boost` reads from `builder`, but never sends any data back).

- `rollup-boost` does not verify liveness of the above connection in any way.

  - meaning that if `builder` would vanish w/o explicitly closing the connection (which is more likely to happen if the builder is truly remote), then `rollup-boost` would have no idea about that and keep trying to read from the socket that is long since has no remote party to push the flashblocks into it.

- normally, the above issue can be mitigated by 2 means:
   
  - both sides implement websocket ping-pong protocol, with certain deadlines on read/write operations (meaning that if either of the messages if failed to be consumed by remote party, then the connection is considered corrupt, and is closed)
     
  - and additionally, the "meaningful" messages (i.e. _not_ ping/pong ones) are also deadlined (e.g. in case of `flashblocks` we could say that if the websocket connection didn't deliver any payload for longer than a second, then it's broken, should be close - and we should try to set up a new one).

0x416e746f6e:
no, it's not covered there.

#298 addresses different problem, namely making sure that `rollup-boost` does reconnect if it _detects_ that the connection is gone.

present issue is about making sure that `rollup-boost` _can_ detect that the connection is gone when it happens abnormally, i.e. when it was just "cut" w/o `rollup-boost`'s OS receiving FIN or RST packet.

the only reliable way of ensuring that is to regularly poll the ws-connection with ping/pong messages.  in http world, that's what keep-alive packets do w/o the developer having to worry about them.  but this is websockets - no keep-alives here.

see also:  https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.2

0x416e746f6e:
> Is it enough to send some data from `rollup-boost` to detect that the websocket is gone? Or we have to receive data back?

if there is data to be sent from `rollup-boost` we could do that, but then it's hard to tell whether it's really received (b/c there's no ack in websocket, unless explicitly implemented by the protocol).

ping/pong with some meaningful payload bytes could do that

karankurbur:
[...]

PR #310, opened 2025-06-09
Rollup-boost uses ping/pong to test the connection is open

ferranbt:
This PR adds a ping/pong system over the Rollup-boost <-> Builder connection. The context of this PR is in #304.

The inbound websocket service will send ping messages to the builder if after some time interval x has not received any message from the builder. After sending the ping message, it will wait another n time duration to receive the pong message back.

The test still needs more work.


ferranbt:
I checked on tungstine and there was not an internal support to do ping.

ferranbt:
> wondering if this can be handled with a library like fastwebsockets since this is quite a common web socket functionality?

Client side implementations do not have heartbeat implemented by default. In the server side it is enabled, if they get a Ping they reply with a Pong.

PR #314, opened 2025-06-10
Rate limit per app/api-key on the authed endpoint

haardikk21:
Introduces functionality to rate limit on a per app or per API Key basis. Also enables being able to set ip rate limiting to 0 which effectively disables it.

### DIFF
--- a/crates/websocket-proxy/src/rate_limit.rs
+++ b/crates/websocket-proxy/src/rate_limit.rs
@@ -20,38 +20,51 @@ pub enum RateLimitError {
 #[clippy::has_significant_drop]
 pub struct Ticket {
     addr: IpAddr,
+    app: Option<String>,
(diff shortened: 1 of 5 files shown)


crates/websocket-proxy/src/rate_limit.rs
```diff
@@ -71,54 +88,102 @@ impl RateLimit for InMemoryRateLimit {
...
-            warn!(
-                message = "ip counting is not accurate -- unexpected underflow",
-                client = addr.to_string()
-            );
-            0
-        } else {
-            current_count - 1
-        };
+        if let Some(app) = app {
+            let current_count = match inner.active_connections_per_app.get(&app) {
```
SozinM, line 167: , we could also moe this to function

PR #319, opened 2025-06-12
Add Flashblocks RPC overlay

ferranbt:
This PR is an initial draft for the Flashblocks RPC overlay in Reth. It is heavily based on the [Base](https://github.com/base/node-reth/tree/main) implementation.

The PR introduces a new binary crate `flashblocks-rpc` which extends the Reth-sdk to run an extension of the JSON-RPC API that provides information about in-flight Flashblocks. The nodes connect to a websocket endpoint to receive the Flashblocks updates.

Be mindful that this is an initial draft in the sense that there are some followup tasks not directly addressed here that I want to tackle as followup PRs, specially around integration tests with the other crates from the workspace.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -110,15 +110,16 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "1.0.9"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ad451f9a70c341d951bca4e811d74dbe1e193897acd17e9dbac1353698cc430b"
+checksum = "b64ec76fe727969728f4396e3f410cdd825042d81d5017bfa5e58bf349071290"
 dependencies = [
  "alloy-eips",
  "alloy-primitives",
  "alloy-rlp",
  "alloy-serde",
  "alloy-trie",
+ "alloy-tx-macros",
  "auto_impl",
  "c-kzg",
  "derive_more",
@@ -126,7 +127,7 @@ dependencies = [
  "k256",
  "once_cell",
  "rand 0.8.5",
- "secp256k1",
+ "secp256k1 0.30.0",
  "serde",
  "serde_with",
  "thiserror 2.0.12",
@@ -148,9 +149,9 @@ dependencies = [
 
 [[package]]
 name = "alloy-dyn-abi"
-version = "1.1.2"
+version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "18cc14d832bc3331ca22a1c7819de1ede99f58f61a7d123952af7dde8de124a6"
+checksum = "7b95b3deca680efc7e9cba781f1a1db352fa1ea50e6384a514944dcf4419e652"
 dependencies = [
  "alloy-json-abi",
  "alloy-primitives",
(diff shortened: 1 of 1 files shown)


crates/flashblocks-rpc/src/cache.rs
```diff
@@ -0,0 +1,279 @@
...
+        // Convert metadata with error handling
+        let metadata: Metadata = match serde_json::from_value(payload.metadata.clone()) {
+            Ok(m) => m,
+            Err(e) => {
+                return Err(eyre::eyre!("Failed to deserialize metadata: {}", e));
+            }
+        };
+
+        if payload.index == 0 {
+            self.reset();
```
avalonche, line 149: there is a time gap between the node getting the latest block and the first flashblock (triggered by block building instead of fcu without attributes). wondering if the cache should be reset on a new block notification instead

crates/flashblocks-rpc/src/cache.rs
```diff
@@ -0,0 +1,282 @@
...
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct Metadata {
+    pub receipts: HashMap<String, OpReceipt>,
+    pub new_account_balances: HashMap<String, String>, // Address -> Balance (hex)
+    pub block_number: u64,
+}
+
+#[derive(Clone)]
+pub struct FlashblocksCache {
+    inner: Arc<Mutex<FlashblocksCacheInner>>,
```
SozinM, line 38: IMO thin is problematic design, because you lock whole structure when in fact you need only part of the structure.

crates/flashblocks-rpc/src/cache.rs
```diff
@@ -0,0 +1,282 @@
...
+
+    pub fn process_payload(&self, payload: FlashblocksPayloadV1) -> eyre::Result<()> {
+        self.inner.lock().unwrap().process_payload(payload)
+    }
+}
+
+struct FlashblocksCacheInner {
+    chain_spec: Arc<OpChainSpec>,
+    builder: FlashblockBuilder,
+    block: Option<OpBlock>,
```
SozinM, line 72: I propose to introduce dashmaps in here, instead of general mutex
For block we could use lock

crates/flashblocks-rpc/src/flashblocks.rs
```diff
@@ -0,0 +1,153 @@
...
+                        tokio::time::sleep(backoff).await;
+                        // Double the backoff time, but cap at MAX_BACKOFF
+                        backoff = std::cmp::min(backoff * 2, MAX_BACKOFF);
+                        continue;
+                    }
+                }
+            }
+        });
+
+        let cache_cloned = self.cache.clone();
```
SozinM, line 87: probably could remove clone (but i'm not sure)

crates/flashblocks-rpc/src/flashblocks.rs
```diff
@@ -0,0 +1,153 @@
...
+
+fn try_parse_message(bytes: &[u8]) -> eyre::Result<String> {
+    if let Ok(text) = String::from_utf8(bytes.to_vec()) {
+        if text.trim_start().starts_with("{") {
+            return Ok(text);
+        }
+    }
+
+    let mut decompressor = brotli::Decompressor::new(bytes, 4096);
+    let mut decompressed = Vec::new();
```
SozinM, line 129: nit - with_capacity

PR #333, opened 2025-06-17
Calculate payload_id locally

SozinM:


0xOsiris:
> isn't there an existing function in alloy that calculates this?

This was taken from Reth I believe: https://github.com/paradigmxyz/reth/blob/671f0fe566a4906a582027e2781c52b1e2bf3104/crates/optimism/payload/src/payload.rs#L330

This impl could change. imo we should expose this in Reth if we want this logic, and upstream. Also what we are benefiting by including this logic other than incurred latency?

0xKitsune:
Agreed with @0xOsiris, what is the motivation for this change? Is there a reason that we are calculating the payload id locally here?

PR #348, opened 2025-06-23
feat: use generics for builder client  

0xKitsune:
This PR updates `RollupBoostServer` to use generics for `builder_client` instead of `dyn EngineApiExt`. Replacing dynamic dispatch with a concrete generic type removes the runtime cost of vtable lookups and heap allocation, enabling the compiler to use static dispatch instead.


0xForerunner (approved):
Looks good to me!

PR #353, opened 2025-06-25
chore: minor tweaks to websocket-proxy

danyalprout:
#### Description
- Add a label to the received message to indicate what upstream uri they came from, this is useful when running with multiple upstream builders (e.g. w/ op-conductor)
- Handle receiving a close frame from the upstream connection (and reconnect)
- Log a warning if we received binary frames (currently not supported)
- Remove the `integration` feature flag / move integration tests to the tests/ dir
- Migrate from redis-test to testcontainers


crates/websocket-proxy/src/registry.rs
```diff
@@ -30,6 +30,7 @@ impl Registry {
                         Ok(_) => {
                             trace!(message = "message sent to client", client = client.id());
                             metrics.sent_messages.increment(1);
+                            metrics.bytes_broadcasted.increment(msg.len() as u64);
```
avalonche, line 33: wondering if its useful here to have an attribute by client (assuming low cardinality of client ids)?

danyalprout: I would expect there to be somewhere in the neighborhood of low thousands of clients (depending on the chain). 

I would probably disable indexing the label on our infra (as this endpoint is public and any number of clients can join), but if it would be useful to ya'll happy to add

PR #357, opened 2025-06-27
chore: cleanup flashblocks logic

0xKitsune:
This PR simplifies and cleans up components of the flashblocks service implementation:

- Simplifies index and base validation logic in `FlashblockBuilder::extend()`.
- Replaces `flat_map` collectors with a single `fold` for more efficient transaction/withdrawal aggregation.
- Deletes the unused `impl From<FlashblocksError> for RpcClientError`.
- Refactors `get_best_payload` to return a single `Result<Envelope>` instead of an `Option` wrapper.
- Minor consistency improvements (naming, formatting).

### DIFF
--- a/crates/rollup-boost/src/flashblocks/service.rs
+++ b/crates/rollup-boost/src/flashblocks/service.rs
@@ -2,7 +2,6 @@ use super::outbound::WebSocketPublisher;
 use super::primitives::{
     ExecutionPayloadBaseV1, ExecutionPayloadFlashblockDeltaV1, FlashblocksPayloadV1,
 };
-use crate::RpcClientError;
 use crate::flashblocks::metrics::FlashblocksServiceMetrics;
 use crate::{
     ClientResult, EngineApiExt, NewPayload, OpExecutionPayloadEnvelope, PayloadVersion, RpcClient,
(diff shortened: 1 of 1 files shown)


crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -193,7 +182,7 @@ pub struct FlashblocksService {
 }
 
 impl FlashblocksService {
-    pub fn new(client: RpcClient, outbound_addr: SocketAddr) -> eyre::Result<Self> {
+    pub fn new(client: RpcClient, outbound_addr: SocketAddr) -> io::Result<Self> {
```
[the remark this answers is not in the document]
SozinM: Is do prefer general approach to errors, because you could always specialize later in case you **really** need to
I think we could ask some more people on that matter, @ferranbt @0xOsiris wdyt?

crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -69,30 +63,28 @@ struct FlashblockBuilder {
...
-            (_, None) => {} // Non-zero index without base is fine
-        }
-
         // Validate the index is contiguous
         if payload.index != self.flashblocks.len() as u64 {
             return Err(FlashblocksError::InvalidIndex);
         }
 
+        // Check base payload rules
+        if payload.index == 0 {
```
ferranbt, line 78: I would like to add a test for this before rewriting it to make sure the migration works as expected.

ferranbt: https://github.com/flashbots/rollup-boost/issues/362

ferranbt: https://github.com/flashbots/rollup-boost/pull/363

PR #358, opened 2025-06-27
chore: simplify `rollup-boost` args

0xKitsune:
This PR cleans up the CLI args structure to improve ergonomics and remove redundant logic:

- Remove the `Commands` subcommand and debug commands. This command seems to be redundant since the debug API is already exposed over HTTP. This also created potential inconsistencies where `execution_mode` could be set via CLI flags but then overridden separately via debug commands.
- Remove the `--flashblocks` boolean flag. Instead, Flashblocks configuration is now optional via `Option<FlashblocksArgs>`. Flashblocks is enabled if any related `--flashblocks-*` flags are specified.
- Rename `Args` to `RollupBoostArgs` for clarity and consistency.

### DIFF
--- a/crates/rollup-boost/src/cli.rs
+++ b/crates/rollup-boost/src/cli.rs
@@ -200,7 +200,6 @@ impl RollupBoostArgs {
                     builder_args.builder_timeout,
                 ));
 
-        // NOTE: clean this up
         let server = Server::builder()
             .set_http_middleware(http_middleware)
             .build(format!("{}:{}", self.rpc_host, self.rpc_port).parse::<SocketAddr>()?)
--- a/crates/rollup-boost/src/cli.rs
+++ b/crates/rollup-boost/src/cli.rs
@@ -1,5 +1,5 @@
 use alloy_rpc_types_engine::JwtSecret;
-use clap::{Parser, Subcommand};
+use clap::Parser;
 use eyre::bail;
 use jsonrpsee::{RpcModule, server::Server};
 use parking_lot::Mutex;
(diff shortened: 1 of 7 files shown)


crates/rollup-boost/src/cli.rs
```diff
@@ -24,10 +23,7 @@ use crate::{
 
 #[derive(Clone, Parser, Debug)]
 #[clap(author, version = get_version(), about)]
-pub struct Args {
-    #[command(subcommand)]
-    pub command: Option<Commands>,
```
0xOsiris, line 29: How can someone startup rollup-boost in a given execution mode without this? e.g. start rollup-boost, but only in dry_run mode? 

I do think one point of confusion here is that the Clap args for execution mode is `dry-run`, while the RPC argument is `dry_run`. Should make these the same naming convention

0xKitsune: There is still the execution_mode cli arg that accomplishes this. 

Previously, there were two ways to set the execution mode on start up, which could cause confusion in the case where the execution mode is set to one setting with the execution_mode flag, and another setting with the debug command commands.

0xOsiris (approved):

0xKitsune:
> This one I would like to keep. I like having a single point of control to turn on the feature. It eliminates the risk of accidentally enabling the feature just by setting a configuration value.

Sounds good, I updated the PR to reintroduce this.

> I have to check first if there is any user using this feature. I like the pattern to ship the service with an easy UX to interact with the internal API endpoints but open to remove it if it is not widely used and the workflow to interact with those endpoints is more automatic than manual.

The main motivation for removing the debug commands was to eliminate the risk of configuring `--execution-mode` via CLI flags and then unintentionally overriding it separately with a debug command. I’m happy to reintroduce it in a way that avoids this inconsistency if you prefer. Historically, we have just used curl to set the execution mode via the Debug API which has been simple, but if you think there’s value in retaining a CLI entry point let me know! Happy to adjust here.

PR #366, opened 2025-07-01
feat: add client ping/pong health checks

danyalprout:
### Description
Add the option to periodically ping clients and disconnect any that don't respond with a pong within a configurable time period. 

Most clients would be disconnected before this, due to lagging by not receiving Flashblocks. However this adds a second line of defense if there's ever a Flashblocks outage.

### DIFF
--- a/crates/websocket-proxy/src/client.rs
+++ b/crates/websocket-proxy/src/client.rs
@@ -1,6 +1,5 @@
 use crate::rate_limit::Ticket;
 use axum::extract::ws::WebSocket;
-use axum::Error;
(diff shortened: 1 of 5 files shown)


SozinM:
Is this supersedes https://github.com/flashbots/rollup-boost/pull/355?

danyalprout:
No, they're healthchecking different connections. #355 does `websocket-proxy -> rollup-boost/builder`, this PR does `websocket-proxy -> client(s)`

avalonche (approved):

Issue #238, opened 2025-05-20
Add flashblocks as a payload source


avalonche:
Right now we distinguish between builder and local l2 as payload sources. However with flashblocks it is possible to fallback to a full builder block instead of a flashblock. It'll be helpful to add tracing and metrics around when this happens.

PR #367, opened 2025-07-02
Add metrics and message log to track number of flashblocks used

SozinM:
Needed to track how good the timings are
closes https://github.com/flashbots/rollup-boost/issues/238

### DIFF
--- a/crates/rollup-boost/src/flashblocks/service.rs
+++ b/crates/rollup-boost/src/flashblocks/service.rs
@@ -228,6 +228,11 @@ impl FlashblocksService {
         // consume the best payload and reset the builder
         let payload = {
             let mut builder = self.best_payload.write().await;
+            let flashblocks_number = builder.flashblocks.len();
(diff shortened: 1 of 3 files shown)


crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -228,6 +228,14 @@ impl FlashblocksService {
         // consume the best payload and reset the builder
         let payload = {
             let mut builder = self.best_payload.write().await;
+            let flashblocks_number = builder.flashblocks.len();
```
avalonche, line 231: I think you can record all the metrics as part of the span and record them there?

SozinM: Good idea, will do it

ferranbt (approved):

SozinM:
@0xKitsune @0xOsiris merged it, because it was pretty basic changes

PR #373, opened 2025-07-10
P2P Flashblocks Propagation

0xForerunner:
Our team was discussing today what the optimal way to distribute flashblocks to consuming nodes is. I feel the websocket proxy method feels a bit forced like pushing web2 apis on something that should really be web3 based. As well, a websocket proxy could be vulnerable do a DDOS attack where some bad actor consumes all available connections on the proxy and is able to take advantage of MEV opportunities without competition.

I've got a _very_ WIP implementation of what distributing flashblocks over the p2p layer could look like instead! Basic flow:
- `rollup-boost` should be in charge of authorizing who has permission to build blocks
- `rollup-boost` signs a message authorizing a builder to build a block with a specific `payload_id`
- signed `Authorization` is included in modified `ForkChoiceUpdated` with attrs call to `builder`
- `builder` receives `Authorization`, begins building blocks and publishes them over the `flashblocks` peering layer as a new libp2p `capability` utilizing the existing peer set
- flashblocks published over the peering layer contain the builders signature along with it's authorization for that `payload_id`
- nodes receive the flashblocks and validate their authorization

I'm posting a draft PR here just in case you're curious to take a look. It's absolutely not ready so please don't leave any comments on implementation. I'm really please with how easy it is to add p2p capabilities on top of reth!


rorupuntou:
u.u

crates/rollup-boost/src/server.rs
```diff
@@ -116,16 +116,23 @@ impl RollupBoostServer<RpcClient> {
         rollup_boost_args: RollupBoostLibArgs,
         probes: Arc<Probes>,
     ) -> eyre::Result<Self> {
-        if rollup_boost_args.flashblocks.flashblocks {
+        if rollup_boost_args.flashblocks_ws.is_some() {
             eyre::bail!(
-                "RpcClient requires flashblocks to be disabled, first check rollup_boost_args.flashblocks.flashblocks == false before calling this constructor"
```
0xOsiris:
```suggestion
                "RpcClient requires flashblocks to be disabled, first check rollup_boost_args.flashblocks.flashblocks_ws == false before calling this constructor"
```

crates/rollup-boost/src/server.rs
```diff
@@ -470,6 +477,19 @@ where
     }
 }
 
+#[rpc(client)]
+pub trait FlashblocksEngineApi {
+    /// When flashblocks is enabled
```
0xOsiris:
```suggestion
    /// When flashblocks p2p is enabled
```

Dockerfile
```diff
@@ -15,7 +15,7 @@ RUN cargo install sccache --version ^0.9
 RUN cargo install cargo-chef --version ^0.1
 
 RUN apt-get update \
-    && apt-get install -y clang libclang-dev
```
0xOsiris: Is this necessary

Cargo.toml
```diff
@@ -27,6 +28,13 @@ tokio = { version = "1", features = ["full"] }
 tracing = "0.1.4"
 tracing-subscriber = { version = "0.3.20", features = ["env-filter", "json"] }
 url = "2.2.0"
+ed25519-dalek = { version = "2", features = ["serde"] }
+blake3 = "1"
+hex = "0.4"
+
+# Reth deps, use 4231f4b to get latest op-alloy
+reth-optimism-payload-builder = { git = "https://github.com/paradigmxyz/reth", rev = "4231f4b" }
```
0xOsiris: These dependencies should be pinned to `v1.9.3`

crates/rollup-boost-types/Cargo.toml
```diff
@@ -15,3 +16,6 @@ op-alloy-rpc-types-engine = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 tracing = { workspace = true }
+thiserror = { workspace = true }
+ed25519-dalek = { version = "2", features = ["serde"] }
+blake3 = "1"                                            # fast hashing for payload IDs
```
0xOsiris: This should be workspace dependency

Dockerfile
0xForerunner: I have this now:
```
RUN apt-get update \
    && apt-get install -y clang libclang-dev gcc

```

Should be fine.

0xOsiris: Yeah I'm asking is this a necessary addition. It was working previously, so what addition made this a necessary change

Cargo.toml
```diff
@@ -4,12 +4,13 @@ resolver = "3"
 members = [
     "crates/rollup-boost",
     "crates/websocket-proxy",
-    "crates/flashblocks-rpc", 
+    "crates/flashblocks-rpc",
```
avalonche: may need to rebase, we removed this crate in favour of the base implementation

PR #375, opened 2025-07-11
Skip "Payload ID mismatch" error if rollup-boost is inactive (not serving FCU with attributes)

SozinM:
We have the problem that if rollup-boost running not serving FCU with attributes and ingesting flashblocks this will cause spam of errors. To fix this i made PayloadId optional and if it's not set (meaning RB inactive) we won't issue an error


SozinM:
Builder playground seems to be broken because of it's recent release

crates/rollup-boost/src/flashblocks/service.rs
```diff
@@ -187,13 +187,15 @@ impl FlashblockBuilder {
...
 
-    // Current payload ID we're processing (set from external notification)
-    current_payload_id: Arc<RwLock<PayloadId>>,
+    /// Current payload ID we're processing. Set from local payload calculation and updated from
+    /// external source.
+    /// None means rollup-boost has not served FCU with attributes yet.
+    current_payload_id: Arc<RwLock<Option<PayloadId>>>,
 
-    // flashblocks payload being constructed
+    /// flashblocks payload being constructed
```
0xOsiris, line 195:
```suggestion
    /// The current Flashblock's payload being constructed.
```

PR #378, opened 2025-07-17
Refactor

malik672:
Cleaned uneccesary clone + remove string allocation


malik672:
Honestly won't it just be  better to use a LRU than this manual impl of the cache system

PR #381, opened 2025-07-24
Add external_state_root and ignore_unhealthy_builders

cody-wang-cb:
Add external_state_root, such that if it's `true`, it will use op-geth for state root calculation

Add ignore_unhealthy_builders flag such that if it's `true`, it won't send builder FCUs when it's unheathy, so EL sync won't be triggered


crates/rollup-boost/src/cli.rs
```diff
@@ -93,6 +93,15 @@ pub struct RollupBoostArgs {
     #[arg(long, env)]
     pub block_selection_policy: Option<BlockSelectionPolicy>,
 
+    /// Should we use the l2 client for computing state root
+    #[arg(long, env, default_value = "false")]
+    pub use_l2_client_for_state_root: bool,
+
+    /// Allow all engine API calls to builder even when marked as unhealthy
+    /// This is default true assuming no builder CL set up
+    #[arg(long, env, default_value = "true")]
```
avalonche, line 102:
```suggestion
    #[arg(long, env, default_value = "false")]
```
default should meant to be false?

crates/rollup-boost/src/server.rs
```diff
@@ -120,8 +130,12 @@ where
             .remove_by_parent_hash(&parent_hash)
             .await;
 
+        let builder_healthy = matches!(self.probes.health(), Health::Healthy);
+
         // async call to builder to sync the builder node
-        if !self.execution_mode().is_disabled() {
+        if !self.execution_mode().is_disabled()
+            && (self.allow_traffic_to_unhealthy_builder || builder_healthy)
```
avalonche, line 137: why is there a builder_healthy check? self.allow_traffic_to_unhealthy_builder should make it redundant?

crates/rollup-boost/src/payload.rs
```diff
@@ -61,6 +61,77 @@ impl OpExecutionPayloadEnvelope {
                 .len(),
         }
     }
+
+    pub fn transactions(&self) -> Vec<Bytes> {
+        match self {
+            OpExecutionPayloadEnvelope::V3(payload) => payload
+                .execution_payload
```
SozinM, line 68: looks scary

crates/rollup-boost/src/server.rs
```diff
@@ -182,20 +196,86 @@ where
...
+            }
+
             // Get payload and validate with the local l2 client
             tracing::Span::current().record("builder_has_payload", true);
             info!(message = "builder has payload, calling get_payload on builder");
 
             let payload = self.builder_client.get_payload(payload_id, version).await?;
-            let _ = self
+
+            if self.use_l2_client_for_state_root == false {
```
SozinM, line 212:
```suggestion
            if !self.use_l2_client_for_state_root {
```

crates/rollup-boost/src/server.rs
```diff
@@ -351,12 +433,19 @@ where
...
             .fork_choice_updated_v3(fork_choice_state, payload_attributes.clone());
+        let builder_healthy = matches!(self.probes.health(), Health::Healthy);
 
         // If execution mode is disabled, return the l2 client response immediately
         if self.execution_mode().is_disabled() {
             return Ok(l2_fut.await?);
         }
 
+        // If traffic to the unhealthy builder is not allowed and the builder is unhealthy,
+        if !self.allow_traffic_to_unhealthy_builder && !builder_healthy {
```
SozinM, line 444: would be happy to see a refactoring, right now there is a lot of conditions, maybe we need to bundle them somehow

crates/rollup-boost/src/server.rs
cody-wang-cb: That's an interesting point, what we thought was that if L2 is able to generate a state root then it's already validated no need to validate again

crates/rollup-boost/src/server.rs
cody-wang-cb: updated

crates/rollup-boost/src/cli.rs
cody-wang-cb: updated

crates/rollup-boost/src/cli.rs
```diff
@@ -93,6 +93,15 @@ pub struct RollupBoostArgs {
     #[arg(long, env)]
     pub block_selection_policy: Option<BlockSelectionPolicy>,
 
+    /// Should we use the l2 client for computing state root
+    #[arg(long, env, default_value = "false")]
+    pub use_l2_client_for_state_root: bool,
```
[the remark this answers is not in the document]
cody-wang-cb: updated

crates/rollup-boost/src/payload.rs
cody-wang-cb: Not sure if there's a better way to structure this though, seems like the only way to fetch the data

crates/rollup-boost/src/tests/unhealthy_builder_traffic.rs
```diff
@@ -0,0 +1,110 @@
...
+    tokio::time::sleep(Duration::from_secs(2)).await;
+
+    // Reset counter after builder should be marked unhealthy
+    *counter.lock().unwrap() = 0;
+
+    // Step 5: Generate blocks - builder should now be unhealthy and with flag=false,
+    // no engine API calls should go to the builder
+    for i in 0..5 {
+        let (_block, block_creator) = block_generator.generate_block(false).await?;
+        println!("Block {}: created by {:?}", i + 1, block_creator);
```
avalonche, line 89: stray print?

cody-wang-cb:
Looks like lint is failing on the rpc side of things unrelated to this PR

crates/rollup-boost/src/server.rs
```diff
@@ -182,20 +192,84 @@ where
...
+            if !self.external_state_root {
+                let _ = self
+                    .l2_client
+                    .new_payload(NewPayload::from(payload.clone()))
+                    .await?;
+
+                return Ok(Some(payload));
+            }
+
+            let fcu_info = self
```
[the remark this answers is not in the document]
cody-wang-cb: Updated

crates/rollup-boost/src/flashblocks/outbound.rs
```diff
@@ -168,15 +169,43 @@ async fn broadcast_loop(
                 }
             }
 
+            // Handle incoming WebSocket messages (including pings)
```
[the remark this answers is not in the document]
cody-wang-cb: i think i got what you meant now, to not check for ping pong but just to use `stream_read.next()`, which works :)

avalonche:
@cody-wang-cb this should be fixed, can you rebase your PR on the latest main?

cody-wang-cb:
@avalonche yup just rebased, looks good now

PR #385, opened 2025-08-01
Always send get_payload to builder

SozinM:
We encountered a problem that if builder does not receive getPayload it will not preserve current executed block to the cache. This change fix this, but adds a little bit of latency in the hot path of flashblocks

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,6 @@
 name: Release
 
 on:
-  workflow_dispatch:
   push:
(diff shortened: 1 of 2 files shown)


avalonche (commented):
why not spawn a new task and use channels to get the result back? it would not increase latency on the hot path

0xOsiris (approved):
This makes sense to me. Seems like a better solution will be caching pre-confirmations in the builder, but should work as a temp patch.

PR #387, opened 2025-08-04
retry FCU if it returns SYNCING

akundaz:
Hacky fix to help out with flashblocks performance


avalonche (commented):
can use a library for this? and generalised function so it can be used across different methods

SozinM:
Propose to close this

PR #389, opened 2025-08-05
wait for newPayload on the builder to complete

akundaz:


SozinM:
What is the idea behind it? 
We won't gain anything by it as i see now

Issue #393, opened 2025-08-18
Add web socket connection with op-node


avalonche:
op-node can connect to rollup-boost over a ws:// connection instead of http:// one.

in the non-flashblocks world, op-node's engine rpc can be set as a ws://url-of-op-geth. So to keep things consistent it would be a nice to have it done the same way against rollup-boost

PR #401, opened 2025-08-28
Add websocket support

SozinM:
closes https://github.com/flashbots/rollup-boost/issues/393

Removed custom traits for APIs, now we are using reth traits directly. This ensures that we will catch all updates.
Fixed issue with vulnerable crate.
Temporary removed flashblocks-rpc from workspace, as it severely outdated.
Bumped reth to 1.7.0


SozinM:
@avalonche 
1. Using pure http proxy breaks websocket
2. When i tried using rpc proxy i found that it's very limited in functionality it supports, there is no easy way to so the same proxying that we were doing with http
3. Authrpc method number is limited, if needed i could just implement all of them (i think i removed couple of them, but could bring them back)

PR #402, opened 2025-08-29
update: flashblocks spec `Metadata` 

julio4:
Update `Metadata` of flashblocks spec as currently used in rollup-boost, node-reth, ...

Potential improvements is define all possible `Receipt` kinds, but it may be unecessary:
- Legacy
- Eip2930
- Eip1559
- Eip7702
- Deposit (Op)


SozinM (approved):
rebase pls

PR #424, opened 2025-10-20
refactor: Enable library usage of rollup-boost

teddyknox:
- **Summary**: Consolidates RPC client configuration into a reusable `ClientArgs` type and refactors CLI, server, and proxy wiring to construct typed RPC/HTTP clients from args. Simplifies `ProxyLayer` to accept prebuilt HTTP clients, adds constructor helpers for `RollupBoostServer`, and improves health/debug startup orchestration.

- **Key changes**
  - Introduce `ClientArgs` with helpers to build clients:
    - `client/rpc.rs`: adds `ClientArgs`, `new_rpc_client`, `new_http_client`, and `define_client_args!` macro to generate `BuilderArgs`/`L2ClientArgs` with consistent flags.
  - Refactor server/CLI construction:
    - `server.rs`: add `RollupBoostServer::<FlashblocksService>::new_from_args` and `RollupBoostServer::<RpcClient>::new_from_args` that validate the `flashblocks` mode and build the appropriate clients; move `ExecutionMode` initialization into `new` and call `update_execution_mode_gauge`.
    - `cli.rs`: build `ClientArgs` from clap args; create `l2_http_client` and `builder_http_client`; start the `DebugServer` from the CLI after constructing `RollupBoostServer`; wire middleware with `ProbeLayer` and the new `ProxyLayer`.
  - Simplify proxy wiring:
    - `proxy.rs`: change `ProxyLayer::new` to accept `HttpClient` instances for L2 and builder rather than URIs/JWT; forward rules unchanged; tests updated to construct clients via `ClientArgs`.
  - Minor cleanups:
    - `flashblocks/mod.rs`: tidy exports.
    - `payload.rs`: make `PayloadTraceContext` and related structs `Debug`.
    - `probe.rs`: return probes from `ProbeLayer::new()` for simpler wiring.
    - `cli.rs`: implement `Default` for `RollupBoostArgs` to aid testing.

- **Breaking changes**
  - `ProxyLayer::new` signature now takes `(l2_client: HttpClient, builder_client: HttpClient)`. All call sites updated.
  - Code constructing `RpcClient`/`HttpClient` should use `ClientArgs` or the new helpers.

- **Operational impact**
  - Behavior is unchanged functionally; startup is cleaner and mode-checked. Health and debug endpoints are initialized from the CLI after server construction.
  - Metrics now record initial execution mode at construction via `update_execution_mode_gauge`.

- **Testing**
  - Updated proxy tests to use `ClientArgs` and the new `ProxyLayer::new`.
  - Added/expanded tests for:
    - Forwarding paths (miner/eth send) and engine passthrough.
    - L2 server recovery and success/failure transitions.
    - Health status behavior under builder/L2 API failures and external state root flow.
  - Server tests now build both RPC and HTTP clients from `ClientArgs` and exercise new constructors.

- **Migration notes**
  - Replace prior `ProxyLayer::new(uri, jwt, ...)` usage with HTTP clients created via:
    - `ClientArgs::new_http_client(PayloadSource::L2|Builder)`
  - Prefer constructing `RollupBoostServer` via `new_from_args` helpers when possible.

### DIFF
--- a/crates/rollup-boost/src/client/rpc.rs
+++ b/crates/rollup-boost/src/client/rpc.rs
@@ -1,5 +1,6 @@
 use crate::EngineApiExt;
 use crate::client::auth::AuthLayer;
+use crate::client::http::HttpClient as RollupBoostHttpClient;
 use crate::payload::{NewPayload, OpExecutionPayloadEnvelope, PayloadSource, PayloadVersion};
 use crate::server::EngineApiClient;
 use crate::version::{CARGO_PKG_VERSION, VERGEN_GIT_SHA};
@@ -10,6 +11,7 @@ use alloy_rpc_types_engine::{
 };
 use alloy_rpc_types_eth::{Block, BlockNumberOrTag};
 use clap::{Parser, arg};
+use eyre::bail;
 use http::{HeaderMap, Uri};
 use jsonrpsee::core::async_trait;
 use jsonrpsee::core::middleware::layer::RpcLogger;
@@ -101,7 +103,7 @@ impl From<RpcClientError> for ErrorObjectOwned {
 ///
 /// - **Engine API** calls are faciliated via the `auth_client` (requires JWT authentication).
 ///
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct RpcClient {
     /// Handles requests to the authenticated Engine API (requires JWT authentication)
     auth_client: RpcClientService,
@@ -376,8 +378,57 @@ impl EngineApiExt for RpcClient {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ClientArgs {
+    /// Auth server address
+    pub url: Uri,
+
+    /// Hex encoded JWT secret to use for the authenticated engine-API RPC server.
+    pub jwt_token: Option<JwtSecret>,
+
+    /// Path to a JWT secret to use for the authenticated engine-API RPC server.
+    pub jwt_path: Option<PathBuf>,
+
+    /// Timeout for http calls in milliseconds
+    pub timeout: u64,
+}
+
+impl ClientArgs {
+    fn get_auth_jwt(&self) -> eyre::Result<JwtSecret> {
+        if let Some(secret) = self.jwt_token {
+            Ok(secret)
+        } else if let Some(path) = self.jwt_path.as_ref() {
+            Ok(JwtSecret::from_file(path)?)
+        } else {
+            bail!("Missing Client JWT secret");
+        }
+    }
+
+    pub fn new_rpc_client(&self, payload_source: PayloadSource) -> eyre::Result<RpcClient> {
+        RpcClient::new(
+            self.url.clone(),
+            self.get_auth_jwt()?,
+            self.timeout,
+            payload_source,
+        )
+        .map_err(eyre::Report::from)
+    }
+
+    pub fn new_http_client(
+        &self,
+        payload_source: PayloadSource,
+    ) -> eyre::Result<RollupBoostHttpClient> {
+        Ok(RollupBoostHttpClient::new(
+            self.url.clone(),
+            self.get_auth_jwt()?,
+            payload_source,
+            self.timeout,
+        ))
+    }
+}
+
 /// Generates Clap argument structs with a prefix to create a unique namespace when specifying RPC client config via the CLI.
-macro_rules! define_rpc_args {
+macro_rules! define_client_args {
     ($(($name:ident, $prefix:ident)),*) => {
         $(
             paste! {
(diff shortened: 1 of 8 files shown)


crates/rollup-boost/src/client/rpc.rs
```diff
@@ -376,8 +378,54 @@ impl EngineApiExt for RpcClient {
...
+        } else {
+            bail!("Missing Client JWT secret");
+        }
+    }
+
+    pub fn new_rpc_client(self: &Self, payload_source: PayloadSource) -> eyre::Result<RpcClient> {
+        RpcClient::new(self.url.clone(), self.get_auth_jwt()?, self.timeout, payload_source).map_err(eyre::Report::from)
+    }
+
+    pub fn new_http_client(self: &Self, payload_source: PayloadSource) -> eyre::Result<RollupBoostHttpClient> {
```
SozinM: copilot comments

crates/rollup-boost/src/server.rs
```diff
@@ -53,9 +55,9 @@ pub struct BuilderPayloadResult {
 pub type BuilderResult = Result<BuilderPayloadResult, ErrorObject<'static>>;
 
 #[derive(Clone)]
-pub struct RollupBoostServer<T: EngineApiExt> {
+pub struct RollupBoostServer {
     pub l2_client: Arc<RpcClient>,
-    pub builder_client: Arc<T>,
+    pub builder_client: Arc<dyn EngineApiExt>,
```
akundaz: What's the reason for using a trait object instead?

akundaz: Thanks for the clarification! It makes sense now looking at `new_from_args`. I think I would lean toward the two constructors approach so that you are completely sure at compile-time which kind of server you're running

crates/rollup-boost/src/proxy.rs
```diff
@@ -25,30 +22,15 @@ const FORWARD_REQUESTS: [&str; 6] = [
 
 #[derive(Debug, Clone)]
 pub struct ProxyLayer {
-    l2_auth_rpc: Uri,
-    l2_auth_secret: JwtSecret,
-    l2_timeout: u64,
-    builder_auth_rpc: Uri,
-    builder_auth_secret: JwtSecret,
-    builder_timeout: u64,
+    l2_client: HttpClient,
```
SozinM: I think with this you might broke compatibility with op-batcher
It relies on connection refused error, and if you reuse client you might not get it

SozinM: how to check:
run rollup-boost (in builder playground for example)
turn off op-geth
see with what error it will reply on rpc call. It should be connection refused

teddyknox: Could you elaborate? I'm not seeing the connection between this rollup-boost PR and op-batcher.

I've tried reproducing the error in the way you suggested by running the flashblocks setup in builder-playground (adding+enabling a feature flag to disable the exiting of the entire setup when one container stops), and then stopping op-geth. I see errors in rollup-boost during this, and then when I restart op-geth again I see the errors stop.

My best guess at what could be a problem with this change is that `Layer` is meant to be a service factory, and so if I use the same HttpClient(s) each time `layer(...)` is called, rather instantiating new clients, then that could potentially be problematic. However, in the implementation of `Service<...> for ProxyService`, [we are cloning self](https://github.com/flashbots/rollup-boost/blob/a434815f586cf384cf0ebe38dc976813c8af1124/crates/rollup-boost/src/proxy.rs#L115) on every request, which I believe resolves this.

teddyknox: Ok to verify, I've pushed a [test](https://github.com/flashbots/rollup-boost/pull/424/files#diff-b5e3227f7e0101a37f6f485ea531d53030345dc69ae66715e4fa34a8400c8d46R798) that verifies that all non-engine API calls to rollup-boost should be directly proxied to the l2 client, and the if the l2 server is down, the proxy returns a retryable error. This test passes both before and after the change in the PR.

0xOsiris (approved):
This generally LGTM. Would wait for other approvals

crates/rollup-boost/src/flashblocks/args.rs
```diff
@@ -23,3 +23,12 @@ pub struct FlashblocksArgs {
     #[arg(long, env, default_value = "5000")]
     pub flashblock_builder_ws_reconnect_ms: u64,
 }
+
+/// Default implementation for FlashblocksArgs
+/// This hack for getting default values from the clap field macro params was
+/// inspired by https://github.com/clap-rs/clap/discussions/5264
+impl Default for FlashblocksArgs {
```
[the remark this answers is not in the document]
teddyknox: You're correct to point out that we do not need this default implementation, [we only need a default implementation for the top-level RollupBoostArgs](https://github.com/flashbots/rollup-boost/pull/424/files#diff-48307ed9ac7c8eb41bdd5bd5a71a4655f0b54dbbe469a8b515927bc1e89a19beR183), so I will remove here and for the L2/Builder client args. The reason that needs a Default implementation is because it's nice for a consuming application to be able to reference these defaults when instantiating RollupBoostServer, e.g. Kona does this [here](https://github.com/op-rs/kona/pull/2915/files#diff-9da6553e1534f2d7d65881ea69d60fcd6d553d5c525bbef87816f3b1e5a8e0f7R572). Specifically, there are fields in RollupBoostArgs which are necessary only for instantiating rollup-boost as a standalone which can be left at defaults in library usage configurations.

SozinM (approved):

PR #426, opened 2025-10-29
feat: handle upstream binary messages to allow being run as a middleman

haardikk21:
Changes:
Now you could chain websocket-proxies

same as https://github.com/base/rollup-boost/pull/21


SozinM (approved):
Looks good!

PR #433, opened 2025-11-06
Replace parking_lot Mutex with tokio Mutex to align with kona-node async primitives

teddyknox:
### Motivation
- parking_lot::Mutex is incompatible with certain concurrency primitives used by kona-node.
- We need an async-aware mutex that cooperates with the Tokio scheduler and avoids blocking the runtime.

### Change summary
- Replace parking_lot::Mutex on async code paths with tokio::sync::Mutex.
- Update all call sites to `lock().await` and adjust signatures to be async where necessary.
- Keep tests and pure-sync helpers on `std::sync::Mutex`.

### Scope
- Updated modules: `server`, `health`, `probe`, `debug_api` (runtime paths) now use `tokio::sync::Mutex`.
- Test and harness code (e.g., request trackers, port allocators) use `std::sync::Mutex`.
- Minor refactors around call-sites to make locking async-safe and avoid `.await` inside non-async contexts.

### Incidental behavior adjustments (in service of the migration)
- Execution-mode handling and health checks were tightened to ensure locks are taken asynchronously and health state updates are non-blocking.
- Builder calls (e.g., FCU/newPayload) are gated consistently behind execution mode and health, ensuring we don’t hold sync locks across awaits.

### Testing and verification
- Updated unit/integration tests to use `std::sync::Mutex` for mock state.
- Assertions and harnesses adjusted to reflect non-blocking health updates and builder gating.
- No external API shape changes; endpoints retain existing request/response formats.

### Risk and rollout
- Touches synchronization across core request paths; requires careful soak in staging.
- Look for deadlock regressions, starvation, or unexpected latency under load.

### Follow-ups
- Audit remaining modules (e.g., `flashblocks/inbound.rs`) for any lingering `parking_lot::Mutex` on async paths and migrate to `tokio::sync::Mutex` where applicable.

### DIFF
--- a/crates/rollup-boost/src/client/rpc.rs
+++ b/crates/rollup-boost/src/client/rpc.rs
@@ -474,7 +474,7 @@ pub mod tests {
     use assert_cmd::Command;
     use http::Uri;
     use jsonrpsee::core::client::ClientT;
-    use parking_lot::Mutex;
+    use std::sync::Mutex;
 
     use crate::payload::PayloadSource;
     use alloy_rpc_types_engine::JwtSecret;
@@ -499,7 +499,12 @@ pub mod tests {
             LazyLock::new(|| Mutex::new(HashSet::new()));
         loop {
             let port: u16 = rand::random_range(1000..20000);
-            if TcpListener::bind(("127.0.0.1", port)).is_ok() && CLAIMED_PORTS.lock().insert(port) {
+            if TcpListener::bind(("127.0.0.1", port)).is_ok()
+                && CLAIMED_PORTS
+                    .lock()
+                    .expect("CLAIMED_PORTS poisoned")
+                    .insert(port)
+            {
                 return port;
             }
         }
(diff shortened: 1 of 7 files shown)


crates/rollup-boost/src/client/rpc.rs
```diff
@@ -474,7 +474,7 @@ pub mod tests {
     use assert_cmd::Command;
     use http::Uri;
     use jsonrpsee::core::client::ClientT;
-    use parking_lot::Mutex;
+    use std::sync::Mutex;
```
theochap: nit: We should tokio mutexes here.

```suggestion
    use tokio::sync::Mutex;
```

crates/rollup-boost/src/tests/common/mod.rs
```diff
@@ -30,6 +29,7 @@ use std::collections::HashSet;
 use std::net::TcpListener;
 use std::path::PathBuf;
 use std::str::FromStr;
+use std::sync::Mutex;
```
theochap: We should use tokio mutexes here

```suggestion
use tokio::sync::Mutex;
```

PR #452, opened 2025-12-01
Fix connection issue with timeout

SozinM:
connect_async does not have any timeouts built in, if there is problems with websocket handshake it will hang.


avalonche (commented):
I think these lines shouldn't be needed as the write will be dropped when Ok(()) is returned
```
  if let Err(e) = write.close().await {
      tracing::warn!("Failed to close builder ws connection: {}", e);
  }
```

PR #481, opened 2026-04-07
chore: remove unsafe fallback in calculate_external_state_root

alessandromazza98:
Stacked on top of #480 

This closes best practice 2 of Nethermind security audit

### DIFF
--- a/crates/rollup-boost/src/health.rs
+++ b/crates/rollup-boost/src/health.rs
@@ -13,6 +13,27 @@ use tracing::warn;
 
 use crate::{EngineApiExt, ExecutionMode, Health, Probes};
 
+pub(crate) fn health_status_when_builder_is_ignored(l2_healthy: bool) -> Health {
+    if l2_healthy {
+        Health::Healthy
+    } else {
+        Health::ServiceUnavailable
+    }
+}
+
+pub(crate) fn health_status_when_builder_is_required(
+    l2_healthy: bool,
+    builder_healthy: bool,
+) -> Health {
+    if !l2_healthy {
+        Health::ServiceUnavailable
+    } else if builder_healthy {
+        Health::Healthy
+    } else {
+        Health::PartialContent
+    }
+}
+
 pub struct HealthHandle {
     pub probes: Arc<Probes>,
     pub execution_mode: Arc<Mutex<ExecutionMode>>,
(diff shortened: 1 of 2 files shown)


crates/rollup-boost/src/health.rs
```diff
@@ -63,49 +85,51 @@ impl HealthHandle {
                             .gt(&self.max_unsafe_interval)
                         {
                             warn!(target: "rollup_boost::health", curr_unix = %t, unsafe_unix = %block.header.timestamp, "L2 client - unsafe block timestamp is too old, updating health status to ServiceUnavailable");
-                            self.probes.set_health(Health::ServiceUnavailable);
+                            self.probes
+                                .set_health(health_status_when_builder_is_ignored(false));
                             sleep_until(Instant::now() + self.health_check_interval).await;
                             continue;
-                        } else if self.execution_mode.lock().is_disabled()
-                            || self.execution_mode.lock().is_dry_run()
```
0xForerunner, line 70: Are we sure we want to remove the dry run logic here?

crates/rollup-boost/src/health.rs
```diff
@@ -63,49 +85,51 @@ impl HealthHandle {
...
-                                warn!(target: "rollup_boost::health", curr_unix = %t, unsafe_unix = %block.header.timestamp, "Builder client - unsafe block timestamp is too old updating health status");
-                                self.probes.set_health(Health::PartialContent);
-                            } else {
-                                self.probes.set_health(Health::Healthy);
-                            }
-                        }
-                        Err(e) => {
-                            warn!(target: "rollup_boost::health", "Builder client - Failed to get unsafe block {} - updating health status", e);
-                            self.probes.set_health(Health::PartialContent);
+                // In enabled and dry-run modes the builder still matters for health.
```
0xForerunner, line 108: I'm not sure I understand the logic here for dry run?

crates/rollup-boost/src/health.rs
alessandromazza98: nethermind audit suggestion says to treat dry_run and enabled in the same way for this specific thing

crates/rollup-boost/src/health.rs
alessandromazza98: essentially for both `enabled` and `dry-run` modes, you need to check both the builder AND the l2 node

PR #484, opened 2026-04-28
fix: restore the old flashblocks-listening env-var

yashvardhan-kukreja:
Signed-off-by: Yashvardhan Kukreja <yashvardhan@oplabs.co>

**Before this PR**

image

**After this PR**

image


yashvardhan-kukreja:
Addressed here - https://github.com/flashbots/rollup-boost/pull/485

